PR #6, opened 2021-11-13
Add Garlicoin COIN target

bradleymackey:
As a bitcoin-derived currency, we can just add a new target.


bigspider:
We are in the process of separating the new Bitcoin app from all the altcoins that rely on the legacy Bitcoin app as a library.
This repository will only contain the new Bitcoin app, therefore this PR will need to be replicated on the https://github.com/LedgerHQ/app-bitcoin legacy repository (which is up-to-date).

Unfortunately, deploying the new bitcoin-only app and the set of bitcoin-derived altcoins needs to happen at the same time for a clean upgrade; that's planned for the beginning of October (not a hard date; hopefully it's not postponed).
You are welcome to open a PR on the legacy repository above.

PR #7, opened 2021-11-30
Standalone python library

bigspider:
This PR extracts a standalone python library (to be released as `ledger_bitcoin`), based on the python library used so far for testing the new protocol.
It generalizes it by providing a compatibility layer that works for the legacy app, by proxying the requests in the new app's API to the old app's APDUs. The compatibility layer is heavily based on https://github.com/bitcoin-core/HWI, which in turn includes a stripped version of the `btchip-python`.

The compatibility layer is tested with speculos against the binaries for the 1.6.5 app.


bitcoin_client/setup.cfg
```diff
@@ -0,0 +1,30 @@
+[metadata]
+name = ledger_bitcoin
+version = 0.0.1
```
lpascal-ledger, line 3: Is there a link with the `__version__` variable in `ledger_bitcoin/btchip/__init__.py`? Version don't match.
If expected, what is the purpose of the latter variable?

bitcoin_client/tests/test_get_master_fingerprint_legacyapp.py
```diff
@@ -0,0 +1,6 @@
+from ledger_bitcoin import Client
+from conftest import SpeculosGlobals
```
lpascal-ledger, line 2: I need this line to be
```
from .conftest import SpeculosGlobals
```
... to work.

bitcoin_client/tests/README.md
```diff
@@ -0,0 +1 @@
+These tests verify the compatibility layer of the bitcoin_client when ran against the legacy app (v1.6.5) using the LegacyClient.
```
lpascal-ledger, line 1: Tests needs to be run from the `tests/` directory, that's worth a notice in the `README`, or using absolute paths in order to properly find `JSON` resources.

bitcoin_client/tests/test_get_master_fingerprint_legacyapp.py
lpascal-ledger: I had to make this change because I ran the tests from the parent directory. No need if they are trigger from this `tests/` directory.

bitcoin_client/setup.cfg
bigspider: The btchip folder is (extracted from) this legacy library https://github.com/LedgerHQ/btchip-python
So this is just the version of that library. Since it's a subpackage, I suppose it does no harm to keep __version__ there, if only for historical info?

bitcoin_client/tests/README.md
bigspider: Good suggestion, I instead fixed the path problems and now they are also runnable from `/bitcoin_client`, which is cool to have.
Some work needed to do the same for the main repository tests, but less important in that case. (Also, there are multiple test folders, so not as useful, perhaps).

bitcoin_client/setup.cfg
lpascal-ledger: Oh sorry I didn't get it was a pick from another library, despite being properly documented in the `bitcoin_client/ledger_bitcoin/btchip/README.md`.

That's a meaningful info all right.

PR #13, opened 2022-01-13
Refactor policy script computation, and some size optimizations

bigspider:
Rewritten the code to compute the derived scripts from a policy; the new version is both smaller and uses less stack.

A few other optimizations to reduce the binary size are also included; most notably, avoiding `uint64_t` divisions in `display.c` saves >400 bytes in binary size.

Overall, this PR reduces the binary size from 63880 to 63112 bytes, or from 62.38 kb to 61.63 kb.


lpascal-ledger (approved):
Would have asked for unit-tests on `uint64_t div10(uint64_t n)`, but as a `static`, couldn't do.

PR #17, opened 2022-02-15
Swap support

bigspider:
This implements the app-exchange swap feature for the new protocol (including taproot addresses), while keeping compatibility with the current version in the legacy protocol.
The code for most of the app-exchange features was extracted from the legacy codebase and moved into the `src/swap` folder, while rewriting it to only rely on the new codebase rather than the legacy one.

The swap support during signing, instead, needs to persist in both the legacy and new signing flow. This slightly increased the app size (about 300 bytes), but it still stays below the 62kb threshold, so it won't currently cause a visible increase to users.

The upgrade doesn't break the current tests of app-exchange; a PR to update app-exchange and add the tests for the new signing flow will be made from the branch https://github.com/LedgerHQ/app-exchange/tree/bitcoin2.


src/main.c
```diff
@@ -385,9 +392,39 @@ static void library_main_helper(struct libargs_s *args) {
...
+
+                io_seproxyhal_init();
+                UX_INIT();
+                ux_stack_push();
+
+                USB_power(0);
+                USB_power(1);
+                // ui_idle();
+                PRINTF("USB power ON/OFF\n");
+#ifdef TARGET_NANOX
```
cbouvet-ledger, line 419: could be HAVE_BLE

src/main.c
```diff
@@ -271,7 +271,10 @@ void app_main() {
                             ui_menu_main,
                             &cmd);
 
-            // TODO: make sure the app exits once signing is done during swap
+            if (G_swap_state.called_from_swap && G_swap_state.should_exit) {
```
cbouvet-ledger, line 274: should_exit is only set if called_from_swap no? why do you need both then?

src/ui/menu.c
```diff
@@ -74,12 +74,12 @@ void ui_menu_main() {
         ux_stack_push();
     }
 
-    if (G_coin_config->kind == COIN_KIND_BITCOIN) {
+    if (G_coin_config->bip32_pubkey_version == 0x0488B21E) {  // mainnet
         ux_flow_init(0, ux_menu_main_flow_bitcoin, NULL);
-    } else if (G_coin_config->kind == COIN_KIND_BITCOIN_TESTNET) {
+    } else if (G_coin_config->bip32_pubkey_version == 0x0488B21E) {  // testnet
```
cbouvet-ledger, line 79: can you enter here? looks like the same as the if above

cbouvet-ledger: isnt the testnet 0x043587CF ? also use defines for that please !

src/legacy/btchip_helpers.c
```diff
@@ -211,23 +211,23 @@ void btchip_write_u32_le(unsigned char *buffer, unsigned long int value) {
...
-
-void btchip_compute_checksum(unsigned char* in, unsigned short inlen, unsigned char * output) {
-    unsigned char checksumBuffer[32];
-    cx_hash_sha256(in, inlen, checksumBuffer, 32);
-    cx_hash_sha256(checksumBuffer, 32, checksumBuffer, 32);
-
-    PRINTF("Checksum\n%.*H\n",4,checksumBuffer);
-    os_memmove(output, checksumBuffer, 4);
-}
+// void btchip_public_key_hash160(unsigned char *in, unsigned short inlen,
```
cbouvet-ledger, line 214: why comment if dead code?

src/main.c
cbouvet-ledger: oh yes alright, defualt value is false ^^

src/legacy/btchip_helpers.c
bigspider: because generally I prefer to touch the `/legacy` folder as little as possible (that's the old codebase). The entire folder will be removed some time this year, anyway :)

src/ui/menu.c
bigspider: Oops, good catch. Fixed in 130a71bec (and also updated the :copyright: notice, while I was there)

src/main.c
bigspider: Leaving as is for now, as it's how the app-boilerplate does

src/ui/menu.c
bigspider: Added tests to catch such bloopers in 7d0f0cfa

PR #19, opened 2022-02-22
[ci] code scanning with SonarCloud

jibeee:


sonar-project.properties
```diff
@@ -0,0 +1,12 @@
+sonar.projectKey=LedgerHQ_app-bitcoin-new
+sonar.organization=ledger
+
+# This is the name and version displayed in the SonarCloud UI.
+sonar.projectName=app-bitcoin-new
+sonar.projectVersion=2.0.2
```
bigspider, line 6: is it necessary to put a version number?
Current version is 2.0.3, and anyway I can imagine this will get unaligned often.

jibeee: It is not necessary. I deleted it in https://github.com/LedgerHQ/app-bitcoin-new/pull/19/commits/3e5faa8ba2c30733504edec3e109ba12f7d9dd33.

PR #30, opened 2022-03-24
allow signing of coinstake transactions

backpacker69:
in order to enable minting from ledger, we need to be able to sign coinstake transactions, coinstake is transaction that allows zero outputs and negative fees (called rewards)

### DIFF
--- a/src/legacy/btchip_apdu_hash_input_finalize_full.c
+++ b/src/legacy/btchip_apdu_hash_input_finalize_full.c
@@ -47,6 +47,7 @@ static bool check_output_displayable() {
     unsigned char amount[8], isOpReturn, isP2sh, isNativeSegwit, j,
         nullAmount = 1;
     unsigned char isOpCreate, isOpCall;
+    bool isRecognizedOutputScript = false;
 
     for (j = 0; j < 8; j++) {
         if (btchip_context_D.currentOutput[j] != 0) {
@@ -70,16 +71,17 @@ static bool check_output_displayable() {
     isOpCall =
         btchip_output_script_is_op_call(btchip_context_D.currentOutput + 8,
           sizeof(btchip_context_D.currentOutput) - 8);
-    if ((((G_coin_config->kind == COIN_KIND_QTUM || G_coin_config->kind == COIN_KIND_HYDRA) &&
-         !btchip_output_script_is_regular(btchip_context_D.currentOutput + 8) &&
-         !isP2sh && !(nullAmount && isOpReturn) && !isOpCreate && !isOpCall) ||
-        (!(G_coin_config->kind == COIN_KIND_QTUM || G_coin_config->kind == COIN_KIND_HYDRA) &&
-         !btchip_output_script_is_regular(btchip_context_D.currentOutput + 8) &&
-         !isP2sh && !(nullAmount && isOpReturn))) &&
-       (!(((G_coin_config->kind == COIN_KIND_PEERCOIN) &&
-         !btchip_output_script_is_regular(btchip_context_D.currentOutput + 8) &&
-         !isP2sh && !(nullAmount && isOpReturn) &&
-         nullAmount)))) {
+    if (btchip_output_script_is_regular(btchip_context_D.currentOutput + 8) ||
+        isP2sh || !(nullAmount && isOpReturn)) {
+        isRecognizedOutputScript = true;
+    } else if (G_coin_config->kind == COIN_KIND_QTUM || G_coin_config->kind == COIN_KIND_HYDRA) {
+        if (isOpCreate || isOpCall)
+            isRecognizedOutputScript = true;
+    } else if (G_coin_config->kind == COIN_KIND_PEERCOIN) {
+        if (nullAmount)
+            isRecognizedOutputScript = true;
+    }
+    if (!isRecognizedOutputScript) {
         PRINTF("Error : Unrecognized output script");
         THROW(EXCEPTION);
     }
--- a/src/legacy/btchip_apdu_hash_input_finalize_full.c
+++ b/src/legacy/btchip_apdu_hash_input_finalize_full.c
@@ -70,12 +70,16 @@ static bool check_output_displayable() {
     isOpCall =
         btchip_output_script_is_op_call(btchip_context_D.currentOutput + 8,
           sizeof(btchip_context_D.currentOutput) - 8);
-    if (((G_coin_config->kind == COIN_KIND_QTUM || G_coin_config->kind == COIN_KIND_HYDRA) &&
+    if ((((G_coin_config->kind == COIN_KIND_QTUM || G_coin_config->kind == COIN_KIND_HYDRA) &&
          !btchip_output_script_is_regular(btchip_context_D.currentOutput + 8) &&
          !isP2sh && !(nullAmount && isOpReturn) && !isOpCreate && !isOpCall) ||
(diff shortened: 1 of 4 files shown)


peerchemist:
@bigspider hello, can you please review this PR?

src/legacy/btchip_apdu_hash_input_finalize_full.c
```diff
@@ -70,12 +70,16 @@ static bool check_output_displayable() {
...
          !btchip_output_script_is_regular(btchip_context_D.currentOutput + 8) &&
          !isP2sh && !(nullAmount && isOpReturn) && !isOpCreate && !isOpCall) ||
         (!(G_coin_config->kind == COIN_KIND_QTUM || G_coin_config->kind == COIN_KIND_HYDRA) &&
          !btchip_output_script_is_regular(btchip_context_D.currentOutput + 8) &&
-         !isP2sh && !(nullAmount && isOpReturn))) {
+         !isP2sh && !(nullAmount && isOpReturn))) &&
+       (!(((G_coin_config->kind == COIN_KIND_PEERCOIN) &&
+         !btchip_output_script_is_regular(btchip_context_D.currentOutput + 8) &&
+         !isP2sh && !(nullAmount && isOpReturn) &&
+         nullAmount)))) {
```
bigspider, line 82: The code was quite unreadable even before the PR, now I'm not able to verify it at all.
I suggest restructuring with the something similar to the following:
```C
  bool is_output_script_acceptable = false;
  if (G_coin_config->kind == COIN_KIND_FOO) {
    // TODO
  } else if (G_coin_config->kind == COIN_KIND_BAR) {
    // TODO
  } else {
    // TODO
  }

  if (!is_output_script_acceptable) {
        PRINTF("Error : Unrecognized output script");
        THROW(EXCEPTION);
  }
```

bigspider (changes requested):
Added a suggestion to make the code more readable.

src/legacy/btchip_apdu_hash_input_finalize_full.c
backpacker69: good idea, thank you, i've refactored this statement like you have asked and it did make it much more readable

src/legacy/btchip_apdu_hash_input_finalize_full.c
```diff
@@ -70,12 +71,17 @@ static bool check_output_displayable() {
...
         btchip_output_script_is_op_call(btchip_context_D.currentOutput + 8,
           sizeof(btchip_context_D.currentOutput) - 8);
-    if (((G_coin_config->kind == COIN_KIND_QTUM || G_coin_config->kind == COIN_KIND_HYDRA) &&
-         !btchip_output_script_is_regular(btchip_context_D.currentOutput + 8) &&
-         !isP2sh && !(nullAmount && isOpReturn) && !isOpCreate && !isOpCall) ||
-        (!(G_coin_config->kind == COIN_KIND_QTUM || G_coin_config->kind == COIN_KIND_HYDRA) &&
-         !btchip_output_script_is_regular(btchip_context_D.currentOutput + 8) &&
-         !isP2sh && !(nullAmount && isOpReturn))) {
+    if (btchip_output_script_is_regular(btchip_context_D.currentOutput + 8) ||
+        isP2sh || !(nullAmount && isOpReturn)) {
```
bigspider, line 75: This is incorrect. It would make any output script accepted if `!(nullAmount && isOpReturn)` is `true`, that is if `(nullAmount && isOpReturn)` is `false`, that is either the output is not null, or it is not an OP_RETURN. That is basically _any_ output script.

The negation should be removed.

bigspider (changes requested):

src/legacy/btchip_apdu_hash_input_finalize_full.c
backpacker69: nice catch, all of negations had to be removed and I've accidentally left this one.

src/legacy/btchip_apdu_hash_input_finalize_full.c
```diff
@@ -70,12 +71,17 @@ static bool check_output_displayable() {
...
-         !isP2sh && !(nullAmount && isOpReturn))) {
+    if (btchip_output_script_is_regular(btchip_context_D.currentOutput + 8) ||
+        isP2sh || (nullAmount && isOpReturn)) {
+        isRecognizedOutputScript = true;
+    } else if (G_coin_config->kind == COIN_KIND_QTUM || G_coin_config->kind == COIN_KIND_HYDRA) {
+        if (isOpCreate || isOpCall)
+            isRecognizedOutputScript = true;
+    } else if (G_coin_config->kind == COIN_KIND_PEERCOIN) {
+        if (nullAmount)
+            isRecognizedOutputScript = true;
```
bigspider, line 82: Alright, the change now seems ok with respect to the other coins.

Regarding Peercoin, this is saying that any script output with null amount is now considered valid. Is this intended behavior? Does it really need to be completely unchecked, or can a condition be added with the right script type?
What happens when a random/invalid script is passed? What is shown on the screen for that output, or does the app crash?

bigspider (changes requested):
Looking better, but likely still incomplete.

src/legacy/btchip_apdu_hash_input_finalize_full.c
backpacker69: Great feedback, I've added extra condition to check for empty script, because coinstake output must have empty script. 

Another thing I've noticed, is that current logic (and old one as well) allows for nullAmount if output script considered regular. Should we add extra check with && !nullAmount to first if?

bigspider: I think that's fine, as a zero-value output does not pose any security risk.
The general idea is to whitelist specific supported valid scripts, and only those.

Last question, otherwise I think it's good: have you checked that the app shows something meaningful for those outputs? Maybe something needs to be added in `main_old.c::get_address_from_output_script`?

backpacker69: Thanks for the tip, it now looks much better.

bigspider (changes requested):
How did you not notice that it doesn't compile?

That suggests that you didn't even test the changes you are suggesting, which would be a reason for rejecting it straight away.

backpacker69:
> Collaborator

apologies for missing part in helpers

peerchemist:
@bigspider are we good to merge this? We are kinda waiting for this to start playing with multisignature minting (staking).

bigspider:
Unfortunately, the 2.0.5 release of the app is already under QA ad will be public soon without this change, since it wasn't sufficiently tested (if at all) when the release was made. It will be considered for the next release.

You can sideload the application on NanoS or NanoS+ or test it with the emulator, meanwhile.

peerchemist:
We are sideloading this app indeed. However the process is cumbersome for many people, it would be much more convenient if the repo were updated.

This code is well tested by now.

bigspider:
Sorry for the slow reply, I have been out of office.

We are in the process of separating the new Bitcoin app from all the altcoins that rely on the legacy Bitcoin app as a library.
This repository will only contain the new Bitcoin app, therefore this PR will need to be replicated on the https://github.com/LedgerHQ/app-bitcoin legacy repository (which is up-to-date).

Unfortunately, deploying the new bitcoin-only app and the set of bitcoin-derived altcoins needs to happen at the same time for a clean upgrade; that's planned for the beginning of October (not a hard date; hopefully it's not postponed).
You are welcome to open a PR on the legacy repository above.

PR #33, opened 2022-04-13
Avoid unwanted global space usage on non-NanoS

bigspider:
Using the global space as a replacement for cxram (optimization to save stack on NanoS) was unsafe and caused memory corruption that broke the exchange app in some situations.
This avoids defining the `get_cxram_buffer` when `USE_CXRAM_SECTION` is not defined, avoiding the risk of using the global space without realizing (app won't compile if `get_cxram_buffer` is used, explicitly labeling code that is using it).

### DIFF
--- a/src/common/base58.c
+++ b/src/common/base58.c
@@ -55,12 +55,15 @@ char const BASE58_ALPHABET[] = {
 };
 
 int base58_decode(const char *in, size_t in_len, uint8_t *out, size_t out_len) {
-    // uint8_t tmp[MAX_DEC_INPUT_SIZE] = {0};
-    // uint8_t buffer[MAX_DEC_INPUT_SIZE] = {0};
-
+#ifdef USE_CXRAM_SECTION
     // allocate buffers inside the cxram section; safe as there are no syscalls here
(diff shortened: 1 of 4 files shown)


src/common/base58.c
```diff
@@ -55,12 +55,15 @@ char const BASE58_ALPHABET[] = {
...
     // allocate buffers inside the cxram section; safe as there are no syscalls here
     uint8_t *tmp = get_cxram_buffer();                          // MAX_DEC_INPUT_SIZE bytes buffer
     uint8_t *buffer = get_cxram_buffer() + MAX_DEC_INPUT_SIZE;  // MAX_DEC_INPUT_SIZE bytes buffer
+#else
+    uint8_t tmp[MAX_DEC_INPUT_SIZE] = {0};
+    uint8_t buffer[MAX_DEC_INPUT_SIZE] = {0};
+#endif
+
     memset(tmp, 0, MAX_DEC_INPUT_SIZE);
     memset(buffer, 0, MAX_DEC_INPUT_SIZE);
```
fbeutin-ledger, line 68: Only necessary if USE_CXRAM_SECTION

fbeutin-ledger (approved):

PR #36, opened 2022-05-03
Fix legacy quirks

bigspider:
Some external libraries using the old `btchip` client library (or derived from them) send some apdus with `CLA = 0xF0`, despite that not being supported in the legacy app for a long time. Since version 2.0.3, only apdus with the `CLA = 0xE0` are sent to the legacy path. This change should address those problem for now (until we completely remove the legacy support from this codebase).


btchip (approved):
ack

Issue #46, opened 2022-05-26
Unable to sign transactions containing mix of input script types


achow101:
With 2.0.4, I am unable to sign transactions that contain multiple input scripts types. The result is `0xb007` with the device displaying a warning about external inputs. Git bisect says that the issue was introduced in commit c41c31117ab9bf1a71617b1e4ad7e34b7ca2e186.

achow101:
Example that should work on speculos

```
[...]

PR #48, opened 2022-05-30
Fix signing with external inputs; added corresponding test

bigspider:
A bug introduced in c41c31117ab9bf1a71617b1e4ad7e34b7ca2e186 prevents signing transactions with external inputs.

This fixes it, and adds a corresponding test.

Closes: #46


sgliner-ledger (approved):
Looks good to me :)

PR #53, opened 2022-06-07
All nano on ci

sgliner-ledger:
Add compilation and tests for Nano X and Nano S+


bigspider (commented):
Great job, very clean! :+1: 

Only a little nit, and it would be nice to generalize a little the `parse_signing_events` function so that it will work with arbitrary screen sizes.

Issue #40, opened 2022-05-10
Use 64-byte Schnorr signatures by default


bigspider:
When the hashtype is the default one (`0`), it would be more appropriate to return 64-byte signatures that omit the hashtype.
Currently, we are instead using 0x01 for the hashtype, that is appended to the signature (resulting in 65-byte signatures).

Details in [BIP-0341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#taproot-key-path-spending-signature-validation) and [BIP-0342](https://github.com/bitcoin/bips/blob/master/bip-0342.mediawiki#signature-validation).

Once this is fixed, it might need to slightly adjust the fee-estimation logic on Ledger-Live side as well.

Issue #47, opened 2022-05-27
SIGHASH_DEFAULT is not recognized


achow101:
If a PSBT has inputs with the sighash type field populated, taproot inputs that have SIGHASH_DEFAULT set are rejected with `Only SIGHASH_ALL is currently supported`. However SIGHASH_DEFAULT is the same as SIGHASH_ALL so this should not be rejected.

bigspider:
Good point. We were already looking into this matter for issue #40, as SIGHASH_DEFAULT should produce 64-byte signatures on taproot and save 1 byte; we'll make sure to add support in the next release.

bigspider:
Solved in #56.

PR #56, opened 2022-06-14
Sighash default support

sgliner-ledger:
Add support for SIGHASH_DEFAULT.
Closes: #40, #47


bigspider (approved):
Nice!
Only one nit: in `test_sign_psbt_taproot_1to2_sighash_all`, I'd add checks to make sure that the signature is 65 bytes and the last byte is indeed `0x01` (it's actually implied by the hardcoded sighash... but still good for documentation purposes imho).

sgliner-ledger:
@bigspider I think those checks were already implemented in the test before I renamed it, and I did not remove it, so it should still be there :)

bigspider:
> When do you plan to release this ? Waiting since June to do a taproot transaction

Sorry for the delay; I'm not sure we'll have a release just for this feature before the next big release that is planned for October.
I'm curious what software are you using that adds SIGHASH_DEFAULT to the psbt?

Sending taproot txs works on Ledger Live, and I think it also works with Sparrow Wallet (they added an option for using SIGHASH_ALL).

oliverlj:
Thank you for your answer.

I'm using sparrow wallet with a personal full node. SIGHASH_DEFAULT is still the default in sparrow

image

Tx broadcasted successfully with SIGHASH_ALL.

I do not want to use Ledger Live according to the tos data collecting to manage my satoshis. And i want to use my personal full node.

bigspider:
Thanks for the feedback! Great that you could work around it until we improve support in the next release.

Not supporting SIGHASH_DEFAULT earlier was an oversight on my side, as it's indeed the sane default choice.

PR #57, opened 2022-06-21
Sighash flags support

### LINKED ISSUE #9: Support for SIGHASH flags
Does the new bitcoin app support ANYONECANPAY SigHashType?

sgliner-ledger:
Add support and tests for all sighash types.
Closes #9.
Inputs with SIGHASH_SINGLE and index bigger or equal to the number of outputs are not supported.


src/ui/display.c
```diff
@@ -479,6 +482,15 @@ UX_FLOW(ux_display_unverified_segwit_inputs_flow,
...
 
+// FLOW to warn about segwitv1 inputs with non-default sighash
+// #1 screen: warning icon + "Non default sighash"
+// #2 screen: "continue" button
+// #3 screen: "reject" button
+UX_FLOW(ux_display_nondefault_sighash_flow,
+        &ux_nondefault_sighash_flow_1_step,
+        &ux_display_continue_step,
+        &ux_display_reject_step);
+
```
bigspider, line 493: Maybe we can be a bit more aggressive and show a `ux_display_reject_if_not_sure_step` ("Reject if you're not sure") screen, until we figure out how people will want to use it and how we do it securely.

It is probably necessary to show more info to safely sign in some protocols (e.g. what inputs have different sighash, what is the sighash), but we can figure this out later − the scary warning should suffice for now.

bigspider (changes requested):
Great job!

- Remove the `test_sign_psbt_with_sighash_types.ipynb` file.
- It is safe to ignore the sighash for external inputs (if any). In fact, if external inputs have sighash flags, but you sign all your inputs with ALL/DEFAULT, there is no problem at all. Therefore, `check_sighash` should probably be skipped for external inputs.
Probably worth adding an extensive comment to `check_sighash` on what it is doing in term of business logic: "If any of the internal inputs has a sighash type that is not SIGHASH_DEFAULT or SIGHASH_ALL, we show a warning".

tests/test_sign_psbt_with_sighash_types.py
```diff
@@ -0,0 +1,453 @@
...
+
+    sig0 = result[0][:-1]
+    sig1 = result[1][:-1]
+
+    assert bip0340.schnorr_verify(sighash_bitcoin_core_all_0, pubkey0, sig0) == 0
+    assert bip0340.schnorr_verify(sighash_bitcoin_core_all_1, pubkey1, sig1) == 0
+
+
+@has_automation("automations/sign_with_default_wallet_accept.json")
+def test_sighash_all_output_modfied(client: Client):
```
bigspider, line 87: typo "modfied"

src/handler/sign_psbt.c
```diff
@@ -1788,11 +1888,22 @@ static void sign_segwit_v1(dispatcher_context_t *dc) {
         }
         crypto_hash_update(&sighash_context.header, tmp, 4);
 
+        if (43 != call_get_merkleized_map_value(dc,
+                                                &state->cur.in_out.map,
+                                                (uint8_t[]){PSBT_IN_WITNESS_UTXO},
+                                                1,
+                                                tmp,
+                                                43)) {
```
bigspider, line 1896: I'm a bit uneasy with the assumption here: the PSBT_IN_WITNESS_UTXO is `<64-bit int> <scriptPubKeylen> <scriptPubKey>`, whose length is `8 + 1 + <length of scriptPubKey>` (well, assuming that the scriptPubKey is not bigger than 252 bytes, which should be a safe assumption). We don't currently support anything with <length of scriptPubKey> > 32, but it could actually be smaller (e.g. a P2SH's scriptPubkey is 22 bytes, because it uses a 20-byte hash, so this check would fail for nested segwit, see [here](bip-0141.mediawiki#p2wpkh-nested-in-bip16-p2sh) − not the case here since we are in segwitv1).

P2TR is currently the only transaction type supported in Taproot, so the assumption currently holds.

Since we only use the first 8 bytes, just checking that it's large enough should be more robust:
```
        if (8 > call_get_merkleized_map_value(dc,
                                              &state->cur.in_out.map,
                                              (uint8_t[]){PSBT_IN_WITNESS_UTXO},
                                              1,
                                              tmp,
                                              8 + 1 + MAX_PREVOUT_SCRIPTPUBKEY)) {
```

Note I also change 43 to the formula `8 + 1 + MAX_PREVOUT_SCRIPTPUBKEY`, so it generalizes if in the future we support larger scriptPubKeys by just changing `MAX_PREVOUT_SCRIPTPUBKEY`.

src/handler/sign_psbt.c
```diff
@@ -1788,11 +1788,22 @@ static void sign_segwit_v1(dispatcher_context_t *dc) {
...
+            SEND_SW(dc, SW_INCORRECT_DATA);
+            return;
+        }
+
         // amount
-        write_u64_le(tmp, 0, state->cur.input.prevout_amount);
         crypto_hash_update(&sighash_context.header, tmp, 8);
 
         // scriptPubKey
+        crypto_hash_update_varint(&sighash_context.header, state->cur.in_out.scriptPubKey_len);
```
bigspider, line 1905: Ooops, luckily ANYONECANPAY was disabled :P

bigspider (changes requested):
Mostly some nits.

It would be nice to have tests `test_sighash_segwitv1_sighashxx` as well; the signature is not deterministic there (hardcoding the sighash, but making sure the transaction is correctly sendable).

Maybe we can work on those test in a separate PR, it might be easier to produce those programmatically with bitcoin-core e2e tests first.

src/handler/sign_psbt.c
sgliner-ledger: I refered to the value specified in the BIP341 for this as you can see [here]. But I can definitely change it.(https://bips.xyz/341#:~:text=amount%20(8)%3A%20value,always%2035%20bytes.)

bigspider: You're right; I thought leaving `state->cur.input.prevout_amount` would work, but it is indeed not initialized when processing outputs (unlike when processing inputs!), so it was wrong before.

bigspider:
Closing and merging onto the `experimental` branch.

PR #62, opened 2022-07-21
Refactor UX calls to be blocking; simplify dispatcher

bigspider:
To be reviewed and merged after #61.

The dispatcher was written last year to cope with the fact that the structure of the app in the boilerplate lacks the ability to:
1) interact with the host computer while processing APDUs (i.e., exchanging messages, that is, the `SW_INTERRUPTED_EXECUTION`/`INS_CONTINUE` mechanism of the app);
2) keep state while the user is interacting with the device.

Ultimately, this boils down to limitations of `io_exchange`, that handles (a) sending a message to the host, (b) getting a response, (c) handling UX, and possibly more.

Because of limitation (2), That means that the `ui_display_` calls used to be non-blocking; this forced to keep the APDU state in the global space rather than in the stack.

The first limitation was already addressed some time in the past, but I couldn't find a way of getting rid of the second limitation.
Finally, I managed that in 4db34a0cdad347cbc8d6cb2cae8f4d5f5238f6c0, by extracting from the sdk the (small!) part of `io_exchange` that deals with UX interactions (see `io_ui_process` in `display.c`).

This allows a bigger refactoring by writing APDU handlers as pure C functions using the stack for any internal state, rather than a state machine operating on the `G_command_state` global variable. As a result of this refactoring:

- The code of handlers (especially `sign_psbt`) is substantially simpler and maintainable.
- Further refactorings to simplify the code became possible.
- The complex features of the dispatcher are removed.
- Handlers no longer make any use of globals (outside of few lines related to the _swap_ feature).
- It should now be much easier to write unit tests for the various sub-functions of the handlers, if desired.
- Binary size is down by more than 1 kb.


src/handler/get_wallet_address.c
```diff
@@ -34,88 +35,109 @@
 
 #include "lib/policy.h"
 #include "lib/get_preimage.h"
+#include "lib/get_merkle_leaf_element.h"
 
 #include "get_wallet_address.h"
 #include "client_commands.h"
 
-static void compute_address(dispatcher_context_t *dc);
+bool parse_and_validate_wallet_policy(dispatcher_context_t *dc);
```
sgliner-ledger, line 43: I think this is never used

sgliner-ledger (approved):

src/handler/get_wallet_address.c
bigspider: Removed in fab57724f7d0aa6a1b40fa2bf5566c72275ad15c; while I was there, I also deleted all the `handler_xxx.h` files that became redundant after the refactoring.

PR #64, opened 2022-08-19
Update JS client libraries; update python lib and app code to wallet policies naming

bigspider:
Updates the library to use version 1 of the protocol (from version 2.1.0 of the app).

Unlike the python library, this doesn't try to stay compatible with previous app versions.
It would be possible to extend compatibility to legacy versions of the app; that could be done in a future PR if desired. The library is not currently in use, anyway.

The PR includes some other collateral changes:
- updated Python library and main app's variable names to match the [wallet policies](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2022-May/020423.html)
- updated Python library to use the new protocol as default, and bumped library version
- added tests of the library with the legacy app to the CI, that was missing


sgliner-ledger (approved):
LGTM :)

PR #77, opened 2022-10-17
Swap fixes

bigspider:
Swap depends on GET_EXTENDED_PUBKEY/GET_WALLET_ADDRESS, that weren't whitelisted.


src/main.c
```diff
@@ -141,11 +141,18 @@ void app_main() {
...
         PRINTF("\n");
 
-        if (G_swap_state.called_from_swap &&
-            (cmd.ins != SIGN_PSBT && cmd.ins != GET_MASTER_FINGERPRINT)) {
-            PRINTF("Only SIGN_PSBT and GET_MASTER_FINGERPRINT can be called during swap\n");
-            io_send_sw(SW_INS_NOT_SUPPORTED);
-            return;
+        if (G_swap_state.called_from_swap) {
+            if (cmd.cla != CLA_APP) {
+                io_send_sw(SW_CLA_NOT_SUPPORTED);
```
sgliner-ledger, line 146: Are you missing a `return` here ?

sgliner-ledger (approved):

src/main.c
bigspider: Good catch, fixed in f739afc.

PR #80, opened 2022-10-26
Allow registering unusual p2pkh or p2wpkh policies

bigspider:
`is_policy_acceptable` was too restrictive, disallowing certain policies that we instead should allow.

Incidentally, adding a fix to the CI due to the recent changes to the `ledger-app-builder`.

### DIFF
--- a/tests/test_register_wallet.py
+++ b/tests/test_register_wallet.py
@@ -10,92 +10,67 @@ from hashlib import sha256
 import pytest
 
 
+def run_register_test(client: Client, speculos_globals, wallet_policy: WalletPolicy) -> None:
+    wallet_policy_id, wallet_hmac = client.register_wallet(wallet_policy)
+
+    assert wallet_policy_id == wallet_policy.id
+
+    assert hmac.compare_digest(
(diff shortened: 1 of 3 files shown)


tests/test_register_wallet.py
```diff
@@ -244,3 +219,32 @@ def test_register_wallet_not_sane_policy(client: Client):
 
     # TODO: we can probably not trigger stack and ops limits with the current limits we have on the
     # miniscript policy size; otherwise it would be worth to add tests for them, too.
+
+
+@has_automation("automations/register_wallet_accept.json")
+def test_register_unusual_singlesig_accounts(client: Client, speculos_globals):
```
sgliner-ledger, line 225: I don't see test with wsh, is it included in previous tests ?

sgliner-ledger (approved):

tests/test_register_wallet.py
bigspider: This test is for the typical single-signature accounts.
Most multisig and all the miniscript e2e tests use wsh, anyway.

(You could do a single signature account inside wsh, but it's pointless − mostly the goal here is to allow people to register accounts when they have funds locked in weird derivation paths).

Issue #99, opened 2023-01-16
Improve error reporting from the client libraries for easy-to-detect errors


bigspider:
Instead of waiting for an error returned by the app, the client library could perform some sanity checks on the inputs.

Examples:
- xpubs fingerprints must be lowercase;
- the only "hardened" symbol supported is `'`, not `h`;
- more?

harshit0verma:
I want to work on this issue. Can you please assign me this issue ?

bigspider:
Thanks @harshit0verma, please let me know if you have any question.
It would be interesting to do this on both the python and the Rust client libraries, but it should be two separate PRs anyway, feel free to just take one if you want.

harshit0verma:
what are "hardened" symbols ?

bigspider:
> what are "hardened" symbols ?

You can read about it in [BIP-32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki) and applications like [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki).

There are two common notations for the symbol; for example, the same valid BIP-44 path could be:
 ```m/44'/0'/0'/1/12```
or:
 ```m/44h/0h/0h/1/12```

Most implementations of BIP-32 accept both notations.
In this app, the first option is chosen as canonical, and the second one is not supported.

Srg213:
Hello. 
Seeing that this issue is still open, I wanted to add changes in rust client libraries.
For handling error, rust client library has enum `BitcoinClientError` .
@bigspider What would be the proper design- adding a new error type or adding a new `Statusword`?
This repo seems to integrate BIP-32 with ledger, I'll check if there are any more checks on input.

bigspider:
cc @edouardparis, in case you want to chime in :pray:

edouardparis:
In this case a new field in the enum https://github.com/LedgerHQ/app-bitcoin-new/blob/develop/bitcoin_client_rs/src/error.rs#L6
can be added. Something like `InvalidParams` which itself would be an enum with values like `InvalidPolicy` and `InvalidKeyInformation`.

I would prefer to not store String message in error enum, in case we want later to facilitate the usage of the lib on memory restricted hardware.

AdarshRawat1:
Can I work on the Python client library.

SrijanShovit:
Hi @bigspider Is this issue still open for contribution or is there something I can help?

PR #101, opened 2023-01-20
#99 issue : added lowercase check to xpub_fingerprints

harshit0verma:
Added a check to check if the xpub fingerprints are all lowercase.

In _base58.py in bitcoin_client added an if statement in "get_xpub_fingerprint" function to check if all the fingerprints are lowercase

### DIFF
--- a/bitcoin_client/ledger_bitcoin/_base58.py
+++ b/bitcoin_client/ledger_bitcoin/_base58.py
@@ -93,6 +93,8 @@ def get_xpub_fingerprint(s: str) -> bytes:
     """
     data = decode(s)
     fingerprint = data[5:9]
+    if(not fingerprint.islower()):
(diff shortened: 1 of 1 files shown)


bigspider (changes requested):
Thanks for your PR!

As a number of files are copied from [HWI](https://github.com/bitcoin-core/HWI), I prefer to avoid touching them to facilitate any future integration work.

I recommend handling errors in the the various `client_*.py` files, instead.
It might make sense to add a new error type in the `exception` submodule if needed (as the errors we're handling are not really "device errors"). Although, in this case python's built-in `ValueError` might be appropriate.

Please make sure to capitalize the error messages (and generally follow the existing style when possible).

bigspider:
Replaced by #106.

PR #102, opened 2023-01-23
More permissive xpub exports

bigspider:
This PR:
- Adds `m/45'/coin_type'/account'` to the allowed paths for pubkey export
- Always allows unhardened subpaths of allowed paths (no point in protecting against it, since they can be computed from the parent)
- Extends the derivation steps limit to 8 instead of 6 (at least one use case is [blind xpubs](/mflaxman/blind-xpub)).


mflaxman:
Love the support for blinded multisig!

It looks like there are a limited subset of derivation paths that can be exported, but after that there is flexibility in BIP32 paths to allow for enhanced privacy.

Last time I was experimenting I could only export the `m/48h/1h/0h/2h` path on testnet:
https://gist.github.com/mflaxman/386990be297e9a595936e8142e815e52

bigspider:
Hey @mflaxman, great to see you here! :)
Your multisig guide was certainly the best material I could find on how to understand multisig, and it certainly influenced a lot the design based on wallet policies in the bitcoin app.

> Love the support for blinded multisig!

Neat idea!

> 
> It looks like there are a limited subset of derivation paths that can be exported, but after that there is flexibility in BIP32 paths to allow for enhanced privacy.
> 
> Last time I was experimenting I could only export the `m/48h/1h/0h/2h` path on testnet: https://gist.github.com/mflaxman/386990be297e9a595936e8142e815e52

The app allows exporting any BIP32 path, except that it requires user confirmation (and gives a big warning) for paths that are non-standard.
The API fails if the `display` parameter is set to `False` for a non-standard path − maybe I didn't document it well-enough.

The idea is to not disrupt the UX for existing use cases (where xpub export was already silent), but harden privacy for all the other paths, so that users who know what they are doing can take advantage of it.

tests/test_get_extended_pubkey.py
```diff
@@ -16,6 +16,11 @@ def test_get_extended_pubkey_standard_nodisplay(client: Client):
         "m/49'/1'/1'/1/3": "tpubDGnetmJDCL18TyaaoyRAYbkSE9wbHktSdTS4mfsR6inC8c2r6TjdBt3wkqEQhHYPtXpa46xpxDaCXU2PRNUGVvDzAHPG6hHRavYbwAGfnFr",
         "m/84'/1'/2'/0/10": "tpubDG9YpSUwScWJBBSrhnAT47NcT4NZGLcY18cpkaiWHnkUCi19EtCh8Heeox268NaFF6o56nVeSXuTyK6jpzTvV1h68Kr3edA8AZp27MiLUNt",
         "m/86'/1'/4'/1/12": "tpubDHTZ815MvTaRmo6Qg1rnU6TEU4ZkWyA56jA1UgpmMcBGomnSsyo34EZLoctzZY9MTJ6j7bhccceUeXZZLxZj5vgkVMYfcZ7DNPsyRdFpS3f",
+        # support up to 7 steps
```
sgliner-ledger, line 19: You could add a test with 8 steps :)

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 Dates are in `dd-mm-yyyy` format.
 
+## [2.1.1] - 23-01-2023
+
+### Changed
+
+- Allow xpub exports at the `m/45'/coin_type'/account'` derivation paths.
+- Allow up to 8 derivation steps for BIP-32 paths (instead of 6).
```
sgliner-ledger, line 15: Maybe you could add that requirements for address index were loosen and checks for change address to be either 1 or 0 were removed :)

sgliner-ledger (approved):

tests/test_get_extended_pubkey.py
bigspider: Oops, initially it was 7. Will fix!

CHANGELOG.md
bigspider: Added a more generic explanation.

tests/test_get_extended_pubkey.py
bigspider: Fixed!

PR #106, opened 2023-01-25
issue 99 fingerprint xpub check and hardened variable check

harshit0verma:
#99  

Modified client_legacy.py for #99 issue.

Added a check for xpub fingerprints to be lowercase .
added a check yo see path only supports ' and not h as hardened variables .


bigspider:
Thanks for your PR. Unfortunately it's incorrect: only the new client supports wallet policies, so adding those checks in `client_legacy.py` is not the right solution.

Please don't make a new PR without spending a bit more time to understand this repository.

(FWI: you can add commits to an existing PR, no need to create a new one when someone requests changes)

PR #114, opened 2023-02-08
Allow setting PSBT_IN_WITNESS_SCRIPT

landabaso:
This PR adds a method to set the `witnessScript` in the input(s) of a PSBTv2 (BIP 370).
This is necessary to be able to sign transactions with utxos having `wsh()` or `sh(wsh())`.


bigspider:
Thanks @landabaso!

PR #116, opened 2023-02-10
JS client: Allow importing from bitcoinjs-lib

landabaso:
* Change the type of the amount from `Buffer` to `number` in the methods `setInputWitnessUtxo` and `getInputWitnessUtxo` in the Javascript client code.
* Add a new method in the Javascript client code to import [BitcoinJS PSBT objects](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/ts_src/psbt.ts):

Example Usage:

```javascript
  const ledgerSignatures = await app.signPsbt(
    new PsbtV2().fromBitcoinJS(bjsPsbt),
    ledgerPolicy,
    ledgerPolicyHmac
  );

```

Please note that the method `fromBitcoinJS` has been tested for `wsh(SCRIPT)` expressions, where `SCRIPT` is a miniscript expression (including `older` and `after` policies). However, it does not support Taproot inputs in the PSBT and will throw an error if attempting to import such PSBTs.

### DIFF
--- a/bitcoin_client_js/src/lib/psbtv2.ts
+++ b/bitcoin_client_js/src/lib/psbtv2.ts
@@ -1,5 +1,7 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 
+import bjs from 'bitcoinjs-lib';
+
 import {
   BufferReader,
   BufferWriter,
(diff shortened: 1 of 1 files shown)


bitcoin_client_js/src/lib/psbtv2.ts
```diff
@@ -371,6 +376,88 @@ export class PsbtV2 {
...
+      if (input.redeemScript)
+        this.setInputRedeemScript(index, input.redeemScript);
+      psbtBJS.data.inputs[index].bip32Derivation.forEach(derivation => {
+        if (derivation.path.substring(0, 2) !== 'm/')
+          throw new Error(`Invalid input bip32 derivation`);
+        const pathArray = derivation.path
+          .replace('m/', '')
+          .split('/')
+          .map(level =>
+            level.match("'") ? parseInt(level) + 0x80000000 : Number(level)
```
bigspider, line 445: Does BitcoinJS always use the `'` symbol for hardened derivation? (that is: it's never `h`)?
Otherwise, `.match("['h]")` handles both; or maybe:
```TypeScript
parseInt(level) + (['h', "'"].includes(level.slice(-1)) ? 0x80000000 : 0)
```

bigspider (approved):
Thanks, this might be useful while we don't have a tested deserialization method!

Only left a small nit.

bitcoin_client_js/src/lib/psbtv2.ts
landabaso: That should be fine. Take a look at the implementation of the "bip32" module in bitcoinjs-lib:

https://github.com/bitcoinjs/bip32/blob/master/ts-src/bip32.ts#L384
Note that bitcoinjs-lib uses a lowercase "m" to denote the master key, and single quotes for hardened paths.

Thank you for accepting the PR, Salvatore. Next week, I'll be releasing a TypeScript descriptors module that will benefit from this change.

landabaso: In any case I'll go ahead and implement those changes to ensure that it won't break if bitcoinjs-lib changes in the future. Better be safe than sorry.

Issue #127, opened 2023-03-07
Don't reject wrapped segwit PSBTs with no non-witness-utxo


bigspider:
Currently, the app is rejecting P2SH-P2WSH or P2SH-P2WPH inputs that have a witness-utxo but no non-witness-utxo.
For consistency, it would be cleaner to accept them with a warning (like for the other segwitV0 inputs).

PR #128, opened 2023-03-08
Allow signing wrapped segwit inputs with missing non-witness-utxo

bigspider:
The app allows signing native segwit inputs with missing non-witness-utxo in the PSBT (since version 2.0.6). However, the logic to detect the condition is based on the input's scriptPubKey, rather than the wallet policy. Therefore, wrapped segwit inputs are outright rejected if the non-witness-utxo is missing.

This PR makes the behavior of the app more consistent: for wrapped segwit inputs, a warning is given if the non-witness-utxo is missing, but signing is otherwise still allowed.

Closes: #127

### DIFF
--- a/src/common/wallet.c
+++ b/src/common/wallet.c
@@ -1876,6 +1876,24 @@ int parse_descriptor_template(buffer_t *in_buf, void *out, size_t out_len, int v
     return parse_script(in_buf, &out_buf, version, 0, 0);
 }
 
+int get_policy_segwit_version(const policy_node_t *policy) {
+    if (policy->type == TOKEN_TR) {
+        return 1;
+    } else if (policy->type == TOKEN_SH) {
+        const policy_node_t *inner =
+            resolve_node_ptr(&((const policy_node_with_script_t *) policy)->script);
+        if (inner->type == TOKEN_WPKH || inner->type == TOKEN_WSH) {
+            return 0;  // wrapped segwit
+        } else {
+            return -1;  // legacy
+        }
+    } else if (policy->type == TOKEN_WPKH || policy->type == TOKEN_WSH) {
+        return 0;  // native segwit
+    } else {
+        return -1;  // legacy
+    }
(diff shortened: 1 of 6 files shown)


tests/automations/sign_with_wallet_missing_nonwitnessutxo_accept.json
```diff
@@ -0,0 +1,21 @@
+{
+  "version": 1,
+  "rules": [
+    {
+      "regexp": "Unverified|Update|or third party|[S]?pend from|Wallet name|Review|Amount|Address|Confirm|Fees",
+      "actions": [
+        ["button", 2, true],
```
sgliner-ledger, line 7: This will require changes to work with Stax :)

sgliner-ledger (approved):

tests/automations/sign_with_wallet_missing_nonwitnessutxo_accept.json
bigspider: Indeed, now I remember that that's why I left the PR as WIP before...

src/common/wallet.c
```diff
@@ -1876,6 +1876,24 @@ int parse_descriptor_template(buffer_t *in_buf, void *out, size_t out_len, int v
     return parse_script(in_buf, &out_buf, version, 0, 0);
 }
 
+int get_policy_segwit_version(const policy_node_t *policy) {
```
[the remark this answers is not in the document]
bigspider: Technically, the segwit version byte is a number between 0 and 16 (except that only versions 0 and 1 are defined at this time).
Also, there are two different ways of returning 0, so I'd still need a comment there.

bigspider:
Rebased on top of #126, and added 2f2cda0bf4a2e4fad42c177e825939bc41c43fba for the Stax automation.

sgliner-ledger (approved):

PR #132, opened 2023-03-13
Update JS library

bigspider:
This does to the JS client library the some change #113 did in the python one.

Moreover, this updates the PSBTV2 class to be able to deserialize a version 0 psbt (and automatically convert it to PsbtV2); the `signPsbt` method is changed so that the `psbt` parameter can be passed as a base64-encoded `string` or as a binary `Buffer`.

This allows integrations to not be exposed to the `PSBTV2` class, and stick to using psbt v0 if preferred (which still has much better support in tooling and libraries).

Even if still unreleased, bumping the package version so that it matches the Python one − it's on par with the client features.

### DIFF
--- a/bitcoin_client_js/src/__tests__/appClient.test.ts
+++ b/bitcoin_client_js/src/__tests__/appClient.test.ts
@@ -319,7 +319,7 @@ describe("test AppClient", () => {
   });
 
   it("can sign a psbt for a taproot script path", async () => {
-    // psbt from test_sign_psbt_singlesig_wpkh_2to2 in the main test suite, converted to PSBTv2
+    // psbt from test_sign_psbt_tr_script_pk_sighash_all in the main test suite, converted to PSBTv2
     const psbtBuf = Buffer.from(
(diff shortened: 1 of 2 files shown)


bitcoin_client_js/src/__tests__/appClient.test.ts
```diff
@@ -296,25 +296,67 @@ describe("test AppClient", () => {
...
+      "hex"
+    ));
+    expect(result[0][1].tapleafHash).toEqual(Buffer.from(
+      "092eda033617e210ee7f7d0e378a404aea1c48b56aa103022becf7746e4700a4",
+      "hex"
+    ));
+
+    // We could test the validity of the signature, but this is already done in the corresponding python test.
+    // Here we're only interested in testing that the JS library returns the correct values.
+    expect(result[0][1].signature.length).toEqual(65); // 64 because it's SIGHASH_ALL and not SIGHASH_DEFAULT
```
sgliner-ledger, line 359: Mismatch between comment and value ?

bitcoin_client_js/src/lib/appClient.ts
```diff
@@ -26,6 +26,42 @@ enum FrameworkIns {
...
+ */
+export class PartialSignature {
+  readonly pubkey: Buffer;
+  readonly signature: Buffer;
+  readonly tapleafHash?: Buffer;
+
+  constructor(pubkey: Buffer, signature: Buffer, tapleafHash?: Buffer) {
+    this.pubkey = pubkey;
+    this.signature = signature;
+    this.tapleafHash = tapleafHash;
```
sgliner-ledger, line 43: I have no idea of how typescript works, but should you check if `tapleafHash` is actually passed as a parameter before assigning it ?

sgliner-ledger (approved):

bitcoin_client_js/src/lib/appClient.ts
bigspider: if not passed, it's implicitly `undefined`; and the `?` itself makes its type `Buffer | undefined`, so that's fine.

bitcoin_client_js/src/__tests__/appClient.test.ts
bigspider: Indeed, and also the comment at the top of the test was wrong. Fixed in d4b6b1df3a96b3d4253fa081d1b4635a5453ea39.

PR #133, opened 2023-03-21
Relax p1 to use it as an optional feature bitmap

bigspider:
For some future use cases, it might be useful to have some additional optional features that can be activated by a client that supports them, but do not otherwise disrupt the normal flow of operations.

The `p1` APDU field is currently unused and forced to be 0.
By relaxing the requirement and ignoring the value of `p1` we guarantee that such optional features can be added without breaking forward-compatibility or forcing a protocol version upgrade (which instead uses the `p2`field).

I also added a test to make sure that a `p2` value higher than the current protocol version is rejected; for the protocol version we prefer explicit opt-in, instead − it would generally be used for more drastic breaking changes to the protocol.

### DIFF
--- a/src/boilerplate/dispatcher.c
+++ b/src/boilerplate/dispatcher.c
@@ -127,7 +127,7 @@ void apdu_dispatcher(command_descriptor_t const cmd_descriptors[],
 
     G_dispatcher_context.read_buffer = buffer_create(cmd->data, cmd->lc);
 
-    if (cmd->p1 != 0 || cmd->p2 > 1) {
+    if (cmd->p2 > 1) {
         io_send_sw(SW_WRONG_P1P2);
         return;
     }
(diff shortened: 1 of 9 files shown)


src/boilerplate/dispatcher.c
```diff
@@ -127,7 +127,7 @@ void apdu_dispatcher(command_descriptor_t const cmd_descriptors[],
 
     G_dispatcher_context.read_buffer = buffer_create(cmd->data, cmd->lc);
 
-    if (cmd->p1 != 0 || cmd->p2 > 1) {
+    if (cmd->p2 > 1) {
```
sgliner-ledger, line 130: nit : If this 1 value corresponds to the protocol version maybe a comment or a define could be nice ? :)

sgliner-ledger (approved):

src/boilerplate/dispatcher.c
bigspider: Good point, added a constant in 4b199cef7deba7789298ed99b78b0ea4898d71de (and rebase to the current `develop`).

PR #141, opened 2023-04-17
Reject policies that are not correctly implemented in certain app versions

bigspider:
As a conservative measure against running miniscript-based wallets with versions of the app affected by the bug fixed in f9673ef7dc5f4b4bcde4f62e44e1c7df4dc803de (versions 2.1.0 and 2.1.1), this PR modifies the python, JS and Rust client to reject operating in any command using a wallet policy if both the following conditions are true:
- the descriptor template contains the `a:` wrapper
- the app version is either `2.1.0` or `2.1.1`.

Incidentally, added a public `getAppAndVersion` method to the JS client library to get the running app name and version, which is available in other clients and could be useful to improve client side support in integrations.

### DIFF
--- a/bitcoin_client_rs/src/wallet.rs
+++ b/bitcoin_client_rs/src/wallet.rs
@@ -252,6 +252,25 @@ impl core::fmt::Display for WalletPubKey {
     }
 }
 
+/// Returns true iff `descriptor_template` contains a `a:` fragment
+pub fn contains_a(descriptor_template: &str) -> bool {
+    let allowed_chars = ['a', 's', 'c', 't', 'd', 'v', 'j', 'n', 'l', 'u']; // miniscript wrappers
+    let mut sequence = String::new();
+
+    for (i, ch) in descriptor_template.chars().enumerate() {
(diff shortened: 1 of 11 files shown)


sgliner-ledger (approved):

bitcoin_client_rs/src/wallet.rs
```diff
@@ -252,6 +252,25 @@ impl core::fmt::Display for WalletPubKey {
     }
 }
 
+/// Returns true iff `descriptor_template` contains a `a:` fragment
+pub fn contains_a(descriptor_template: &str) -> bool {
```
edouardparis, line 256: I think you can use `descriptor_template.contains("a:")` instead

bigspider: but multiple wrappers could be present before the `:`, so e.g. you could hav something like `ac:...`

edouardparis: Ah right ! I am less qualified in miniscript, I did not know you can stack wrappers. Good to go then

PR #143, opened 2023-04-18
Add address test for all miniscript fragments 

bigspider:
Still following up on f9673ef7dc5f4b4bcde4f62e44e1c7df4dc803de, this adds a test that compares the generated addresses with the ones obtained from bitcoin-core for a set of descriptors that covers all miniscript fragments.


bigspider:
Changes from this PR included in #155.

PR #160, opened 2023-05-04
Add guidelines enforcer to CI

bigspider:
Also adds curve and path restrictions to the MAKEFILE.

### DIFF
--- a/src/crypto.c
+++ b/src/crypto.c
@@ -276,8 +276,7 @@ bool crypto_get_compressed_pubkey_at_path(const uint32_t bip32_path[],
                                           uint8_t pubkey[static 33],
                                           uint8_t chain_code[]) {
     struct {
-        uint8_t prefix;
-        uint8_t raw_public_key[64];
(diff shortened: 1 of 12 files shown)


src/crypto.c
```diff
@@ -287,7 +286,6 @@ bool crypto_get_compressed_pubkey_at_path(const uint32_t bip32_path[],
     bool result = true;
     BEGIN_TRY {
         TRY {
-            keydata.prefix = 0x04;  // uncompressed public keys always start with 04
```
sgliner-ledger, line 290: Do we not need this anymore ?

bigspider: The memcpy below now copies the entire 65-byte uncompressed pubkey before (including the prefix), so the code is simpler. Earlier, the first byte was skipped is the memcpy, so I was adding the prefix by hand. Uncompressed pubkeys always start with `0x04`.
The change is to make the code analyzers happy, as casting a `struct` to a `uint8_t *` made them _nervous_.

PR #166, opened 2023-05-29
Compare generated addresses with the expected ones in the client library

bigspider:
Supersedes #141 with a more generic solution: derive addresses from the wallet policy on the client side, and fail if they differ.

The libraries used for client side are generation are externally developed.


bitcoin_client_js/src/lib/appClient.ts
```diff
@@ -404,14 +404,32 @@ export class AppClient {
 
   /* Performs any additional checks on the policy before using it.*/
   private async validatePolicy(walletPolicy: WalletPolicy) {
+    // TODO: Once an independent implementation of miniscript in JavaScript is available,
+    // we will replace the checks in this section with a generic comparison between the
+    // address produced by the app and the one computet locally (like the python and Rust
```
sgliner-ledger, line 409: compute`d`

sgliner-ledger (dismissed):

landabaso:
Hi there. I just saw this.
For Javascript, you can use https://bitcoinerlab.com/modules/descriptors
Let me know if you need some assistance. I'm happy to help.

Issue #109, opened 2023-01-30
Improve UX for self-transfers


bigspider:
Currently, when sending coins to an address controlled by the wallet policy, the UX only considers an output a _change_ address if the the BIP-32 ends with `/1/*` (or more generally `/N/*`, for wallet policies containing `<M;N>/*`).

This is consistent with the UX the previous app had (and probably most hardware signers); however, it might be confusing for users, as it might not match their expectations. Moreover, for security, the (internal) output address would have to be first validated on the secure screen, and then one would have double-check that the output is sent to the same address. Users are likely to skip some steps, or performing them insecurely (for example, copying the address on an editor on the same machine they are using to send the transaction).

Possible solutions:
- Consider any internal outputs change; for a transaction where all the inputs/outputs are internal, one could have an explicit UX showing it as "self-transfer", as the sent amount is 0 (apart from the fees). Some care needs to be taken to check to handle the case where there are also external inputs.
- Keep the same UX flow, but explicitly label outputs with `/0/*` or `/M/*` as "your own" in the UX.

PR #173, opened 2023-06-19
Add special wording for self-transfers

bigspider:
Closes: #109


src/ui/display_nbgl.c
```diff
@@ -204,6 +212,37 @@ void ui_accept_transaction_flow(void) {
     nbgl_refresh();
 }
 
+void ui_accept_selftransfer_flow(void) {
```
sgliner-ledger, line 215: I think here we could try to avoid code duplication by moving all of the code except the tagValue filling to a dedicated function :)

sgliner-ledger (dismissed):

src/ui/display_nbgl.c
bigspider: Solved in 33f2e13 by adding an `is_self_transfer` parameter, so I could instead delete both `ui_validate_selftransfer` and `ui_accept_selftransfer_flow`.

src/ui/display.h
```diff
@@ -132,7 +132,12 @@ bool ui_validate_output(dispatcher_context_t *context,
                         const char *coin_name,
                         uint64_t amount);
 
-bool ui_validate_transaction(dispatcher_context_t *context, const char *coin_name, uint64_t fee);
+bool ui_validate_transaction(dispatcher_context_t *context,
+                             const char *coin_name,
+                             uint64_t fee,
+                             bool is_self_transfer);
+
+bool ui_validate_selftransfer(dispatcher_context_t *context, const char *coin_name, uint64_t fee);
```
sgliner-ledger, line 140: You can drop this

sgliner-ledger (dismissed):

sgliner-ledger (approved):

PR #182, opened 2023-07-07
Improve and refactor logic for standard wallet checking

bigspider:
Fixes a bug where some non-default policies would be accepted as default.

Meanwhile, also refactors and simplifies the logic to verify if a policy is standard (reducing code duplication between 
`get_wallet_address` and `sign_psbt`, and updates the docs on wallet policies to the latest (testnet) deployed version.


src/handler/lib/policy.c
```diff
@@ -1285,25 +1286,127 @@ __attribute__((noinline)) int get_wallet_internal_script_hash(
...
+            if (((const policy_node_tr_t *) descriptor_template)->tree != NULL) {
+                return false;
+            }
+
+            kp = ((const policy_node_tr_t *) descriptor_template)->key_placeholder;
+            purpose = 86;  // standard single-key P2TR
+            break;
         default:
-            return -1;
+            return false;
```
sgliner-ledger, line 1323: Looks a bit weird to me to have a `static int` function returning `false`

sgliner-ledger (approved):

src/handler/lib/policy.c
bigspider: Oops, good catch, the would return 0, which is a bug as that could technically be a valid purpose.
Remnant of a previous function returning `bool` (but how does that _not_ give a warning? Aaaah, C..)!

bigspider: Fixed (and rebased).

src/handler/lib/policy.c
```diff
@@ -1285,25 +1286,127 @@ __attribute__((noinline)) int get_wallet_internal_script_hash(
...
+
+    return purpose;
+}
+
+bool is_wallet_policy_standard(dispatcher_context_t *dispatcher_context,
+                               const policy_map_wallet_header_t *wallet_policy_header,
+                               const policy_node_t *descriptor_template) {
+    // Based on the address type, we set the expected bip44 purpose
+    int bip44_purpose = get_bip44_purpose(descriptor_template);
+    if (bip44_purpose < 0) {
```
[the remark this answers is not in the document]
bigspider: Having fixed the bug above, that can't happen; only a value < 0 signals an error.

src/handler/get_wallet_address.c
```diff
@@ -133,88 +132,21 @@ void handler_get_wallet_address(dispatcher_context_t *dc, uint8_t protocol_versi
...
             return;
         }
 
-        // check if derivation path is indeed standard
-
-        // Based on the address type, we set the expected bip44 purpose
-        int bip44_purpose = get_bip44_purpose(address_type);
-
-        if (key_info.master_key_derivation_len != 3) {
+        if (wallet_header.name_len != 0) {
```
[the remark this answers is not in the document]
bigspider: Technically, I consider the wallet policy as just the "descriptor template" and the keys; so I prefer to keep that out of the arguments for any function that works with wallet policies.
Also, not a security issue even if this check were to be skipped.

sgliner-ledger (approved):

PR #183, opened 2023-07-11
bump bitcoin-0.30 and miniscript-10.0

edouardparis:
`psbt::serialize::{Serialize, Deserialize}` are private traits I re implemented them in order to serialize psbt key/value pairs.
Hopefully one day rust-bitcoin support PSBTv2

### DIFF
--- a/bitcoin_client_rs/tests/client.rs
+++ b/bitcoin_client_rs/tests/client.rs
@@ -2,9 +2,9 @@ mod utils;
 use std::str::FromStr;
 
 use bitcoin::{
-    consensus::encode::deserialize,
-    hashes::hex::{FromHex, ToHex},
-    util::{bip32::DerivationPath, psbt::Psbt},
+    bip32::DerivationPath,
+    hashes::{hex::FromHex, Hash},
+    psbt::Psbt,
 };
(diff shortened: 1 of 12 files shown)


edouardparis:
I want the code to look like exactly like `bitcoin::psbt::serialize` module in order to reduce risk of errors, also rust-bitcoin psbt module exposes publicly only a few methods for few types. In order to have the PSBTv2 transition I needed to have access to a very low level of the PSBT mapping.

darosior:
Oh, gotcha. I overlooked the v2 part.

bitcoin_client_rs/tests/client.rs
```diff
@@ -310,15 +310,50 @@ async fn test_sign_psbt() {
...
+                                    .unwrap(),
+                                hex::encode(sig.to_vec())
+                            );
+                        }
+                        _ => {}
+                    }
+                }
+            }
+        }
+
```
bigspider, line 356: There's identical code in lines 362:395, perhaps it can be in a function?
Even an inner function might do.

bigspider (changes requested):
Thank you!
Only commented on a couple of nits.

bigspider (approved):

PR #185, opened 2023-07-17
Improve pubkey memory management

bigspider:
Because of legacy code, the app until now used serialized (bip58-encoded) pubkeys everywhere. That's less efficient for no good reason, and requires more memory.

Also took a chance to remove a couple of old TODOs and add a few comments.

### DIFF
--- a/src/handler/get_extended_pubkey.c
+++ b/src/handler/get_extended_pubkey.c
@@ -20,6 +20,7 @@
 #include "boilerplate/io.h"
 #include "boilerplate/dispatcher.h"
 #include "boilerplate/sw.h"
+#include "../common/base58.h"
 #include "../common/bip32.h"
 #include "../commands.h"
 #include "../constants.h"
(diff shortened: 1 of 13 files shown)


src/handler/get_extended_pubkey.c
```diff
@@ -141,27 +142,41 @@ void handler_get_extended_pubkey(dispatcher_context_t *dc, uint8_t protocol_vers
...
+    int pubkey_str_len = base58_encode((uint8_t *) &pubkey_check,
+                                       sizeof(pubkey_check),
+                                       pubkey_str,
+                                       sizeof(pubkey_str));
+    if (pubkey_str_len != 111 && pubkey_str_len != 112) {
+        PRINTF("Failed encoding base58 pubkey\n");
         SEND_SW(dc, SW_BAD_STATE);
         return;
     }
+    pubkey_str[pubkey_str_len] = 0;
```
sgliner-ledger, line 169: For clarity I would have replaced this 0 with `\0`

sgliner-ledger (approved):

src/handler/get_extended_pubkey.c
bigspider: I always did `= 0` in the codebase, so leaving the same for consistency; can be changed everywhere in a separate PR if desired.

PR #191, opened 2023-08-02
do not initialize Decimal from float

fametrano:
this avoid warnings and prefer explicit proper initialization


bigspider (approved):
Thank you!

PR #199, opened 2023-10-31
Use image compiled the latest bitcoin-core instead of a custom fork

bigspider:
Now that bitcoin-core has full miniscript support, we no longer need to use a custom branch for e2e tests.

### DIFF
--- a/.github/workflows/Dockerfile
+++ b/.github/workflows/Dockerfile
@@ -1,14 +1,30 @@
 # An image derived from ledgerhq/speculos but also containing the bitcoin-core binaries
+# compiled from the master branch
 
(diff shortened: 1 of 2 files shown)


bigspider:
TBC once bitcoin v26.0 is published [here](https://bitcoincore.org/en/releases/); needs to be updated in the Dockerfile.

bigspider:
On second thought, a8fce7cdf4fed3c04b8805796df9e58f8cec504a compiles bitcoin-core from github instead of using a specific tagged release.

The chance of changes to the bitcoin-core wallet that break the e2e tests is quite low, and if that were to happen, we probably want to know as soon as possible, rather than waiting for the next tagged release of bitcoin-core.

PR #203, opened 2023-11-10
Swap improvements

sgliner-ledger:


src/boilerplate/sw.h
```diff
@@ -36,6 +36,11 @@
  */
 #define SW_WRONG_DATA_LENGTH 0x6A87
 
+/**
+ * Status word for fail in Swap
+ */
+#define SW_FAIL_SWAP 0x6A8E
+
```
bigspider, line 43: Is this status word also used in other apps for the swap integration?
I would have preferred not to use an SW in the `0x6AXX` range, and maybe use `0x6B00` or something similar. Could also use a subrange like `0x6BXX` for swap, so that different error conditions could be detected appropriately.

src/handler/sign_psbt.c
```diff
@@ -649,8 +650,8 @@ init_global_state(dispatcher_context_t *dc, sign_psbt_state_t *st) {
     // Swap feature: check that wallet policy is a default one
     if (G_swap_state.called_from_swap && !st->is_wallet_default) {
         PRINTF("Must be a default wallet policy for swap feature\n");
-        SEND_SW(dc, SW_INCORRECT_DATA);
-        return false;
+        SEND_SW(dc, SW_FAIL_SWAP);
+        finalize_exchange_sign_transaction(false);
```
bigspider, line 654: An issue that was indeed already present (and therefore, we could solve in a separate PR) is that `finalize_exchange_sign_transaction` is only called if there is an error in the logic strictly related to swaps, but there are other places where an error could happen.

The right way of fixing this might be to remove all the `SEND_SW` in the code, and instead use the return values to signal/propagate errors.

We could take the change to do a bigger refactoring error reporting more generally, for example with a short error message in the response APDU instead of just a status word. That would help debugging, as people having issues only know that they are getting "INCORRECT_DATA", but there are tens of places where that could be returned...

bigspider (approved):

src/boilerplate/sw.h
sgliner-ledger: It is used by some other apps, but I can change it for sure :) 
I'll go for `0x6B00`

src/handler/sign_psbt.c
sgliner-ledger: +1

bigspider (approved):

PR #217, opened 2024-01-17
Use rust-bitcoin 0.31 and change version for v0.4.0

edouardparis:
Paranoid client still makes a test fail because of missing apdu exchanges in the mock.
 but all tests are working using:
```
cargo test --no-default-features --features async
```


bigspider (approved):
LGTM

PR #223, opened 2024-02-05
Avoid app crashing on certain malformed PSBTs on legacy inputs

bigspider:
When signing for legacy inputs, the app makes the assumption that the witness-utxo field is absent. The condition was not properly enforced during preprocessing, so it would only be detected at signing time.

There, an incorrect handling of the error status word caused the app to crash after returning the error condition.

This fixes both issues.


sgliner-ledger (approved):
At some point it could be nice to handle all `SEND_SW` done from this function I guess :)

Issue #228, opened 2024-02-19
Add status word `0x5515` to the client libraries


bigspider:
The SDK returns this status word when the device goes on auto-lock and an APDU is sent.

bigspider:
After #226, the `SW_SECURITY_STATUS_NOT_SATISFIED` is no longer used by the app. We can therefore redefine `SW_SECURITY_STATUS_NOT_SATISFIED` to equal `0x5515` instead of the current value `0x6982`, since the purpose was the same. In this way, any client wallet that was depending on the cient libraries for the handling of `SW_SECURITY_STATUS_NOT_SATISFIED` will work without needing explicit update.

For compatibility with older app versions, we could map both status words to the same error class in the client libraries.

PR #229, opened 2024-02-19
Handle 0x5515 status word in python client library

bigspider:
Closes: #228


bitcoin_client/ledger_bitcoin/exception/device_exception.py
```diff
@@ -6,8 +6,9 @@
 
 class DeviceException(Exception):  # pylint: disable=too-few-public-methods
     exc: Dict[int, Any] = {
+        0x5515: SecurityStatusNotSatisfiedError,  # returned by sdk in recent versions
```
sgliner-ledger, line 9: It seems a bit confusing to say recent and older without specific versions :sweat_smile:

sgliner-ledger (approved):

bitcoin_client/ledger_bitcoin/exception/device_exception.py
bigspider: Yeah, but unfortunately it's not easy to track down the exact version; not critical, though - at least we fix the default current behavior.

PR #238, opened 2024-03-07
Ragger tests

sgliner-ledger:


tests/test_e2e_miniscript.py
```diff
@@ -83,6 +83,9 @@ def run_test_e2e(wallet_policy: WalletPolicy, core_wallet_names: List[str], rpc:
     result = multisig_rpc.walletcreatefundedpsbt(
         outputs={
             out_address: Decimal("0.01")
+        },
+        options={
+            "changePosition": 1
```
bigspider: Still not a big fan of this; it would be nice to have the change address in a random position.

At least, it would be nice to add a comment explaining why this is here (probably copying the comment in `tests/test_e2e_multisig.py` and `tests/test_e2e_tapscripts.py`; refactoring the common code could be a future PR).

bigspider (changes requested):
This is a great improvement to the test suite! 👏

Some things that would be nice to fix before merging:
- ideally, `pytest` should work without argument, at least if ran from the right place; if not, at least the `README.md` should be up to date with the latest instructions on how to run the tests.
- Before this PR, the test suite had the following optional arguments:
    - `--hid` to run with the real device
    - `--headless` to run without speculos UX
    - `--model` to specify which model
I understand that those are now all overlapping with ragger features, so the options should be removed (and ideally some little explainer added to README.md).

A couple of minor nits in inline comments.

tests/test_e2e_miniscript.py
sgliner-ledger: I can duplicate the tests to test all the possible positions if you prefer :shrug:

bigspider: I think just a one-line comment explaining why it's there (e.g. what breaks without, and why if you know it) would do. Can improve it in a separate future PR.

PR #242, opened 2024-03-13
FIX python exemple

pgrange:
The python exemple code in the README would not work. The reason seems to be the fact that `/**` at the end of a key is not allowed in a policy.

The following patch fixes this issue.


bigspider:
Thanks @pgrange, that was indeed a remnant of the old wallet policies (before version 2.1.0), when the `/**` was moved from the key info to the descriptor template.

Issue #234, opened 2024-02-26
Missing return checks on some hash functions


bigspider:
Some functions that return void call SDK functions that potentially return an error code, which is currently unchecked. The error doesn't ever happen since the parameters of the functions are correct, but we should add a `LEDGER_ASSERT` as defensive programming.

Functions to cover: `crypto_get_checksum`, `crypto_ripemd160`, `crypto_hash160`, `crypto_tr_tagged_hash_init` and `crypto_tr_tapleaf_hash_init`.

PR #249, opened 2024-03-27
Add return value checks for hash functions from the sdk

bigspider:
Closes: #234

### DIFF
--- a/src/crypto.c
+++ b/src/crypto.c
@@ -157,7 +157,7 @@ int bip32_CKDpub(const serialized_extended_pubkey_t *parent,
 
 void crypto_ripemd160(const uint8_t *in, uint16_t inlen, uint8_t out[static 20]) {
     int res = cx_ripemd160_hash(in, inlen, out);
-    LEDGER_ASSERT(res == CX_OK, "Unexpected error in ripemd160 computation");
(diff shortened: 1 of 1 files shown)


src/crypto.c
```diff
@@ -156,14 +156,16 @@ int bip32_CKDpub(const serialized_extended_pubkey_t *parent,
...
+    LEDGER_ASSERT(res == CX_OK, "Unexpected error in ripemd160 computation");
 }
 
 void crypto_hash160(const uint8_t *in, uint16_t inlen, uint8_t out[static 20]) {
     PRINT_STACK_POINTER();
 
     uint8_t buffer[32];
-    cx_hash_sha256(in, inlen, buffer, 32);
+    int res = cx_hash_sha256(in, inlen, buffer, 32);
+    LEDGER_ASSERT(res == CX_SHA256_SIZE, "Unexpected error in sha256 computation");
```
sgliner-ledger, line 168: It could be nice to add the res value in the log message :woman_shrugging:

sgliner-ledger (dismissed):

src/crypto.c
bigspider: Done in f4d38edfcf4a2da5e55ff9813fe7d61abc6677c8.

Might be worth adding that in the boilerplate examples.

Issue #248, opened 2024-03-22
Improve UX wording in transaction signing flow


bigspider:
The current wording `Accept and send` as the final step to approve a transaction never matches the user experience for multisignature wallet policies, where multiple signatures are required and the actions of "signing" and "broadcasting" a transaction are clearly distinct.

Simply `Sign transaction` might be a better choice.

PR #252, opened 2024-04-05
Improve wording of 'Sign transaction' UX flows. 

bigspider:
This changes the wording during transaction signatures:

- The 'Approve' step after each output becomes 'Continue'. It was reported that 'Approve' was confusing as users might think they are already approving the entire transaction. 
- The 👁️"Confirm transaction" screen is removed (except for self-transfers, as there's no output to show in that case)
- The 'Accept and send' becomes 'Sign transaction'. This is more precise, as broadcasting transactions is the responsibility of the software wallets, and it doesn't necessarily happen once the user approves the transaction (particularly, that's confusing when using multisignature or more complex multi-user spending policies).

Before:
txflow-old
After (but without removing the "Confirm transaction" step):
txflow-new

Final:
txflow-new-2

Closes: #248


bigspider:
Before:
txflow-old
After:
txflow-new

blucet-ledger:
@bigspider the screen preceding the fee display shows "Confirm transaction", which seems too generic to be useful for users. What do you think of:
- making it more informative and accurate, eg. "Review fees";
- or removing it altogether?

bigspider:
> @bigspider the screen preceding the fee display shows "Confirm transaction", which seems too generic to be useful for users. What do you think of:
> 
> * making it more informative and accurate, eg. "Review fees";
> * or removing it altogether?

Not opposed to idea; I'd prefer removing it, it does indeed seem redundant.

bigspider:
If there are no outputs (all the money spent is going to a change address), since version `2.1.3` we show a simplified UX:

txflow-selftransfer

In this case it's worth keeping the "Confirm self-transfer" screen, even if we remove it for normal transactions.

bigspider:
As per the discussion above, deleted the 'Confirm transaction' screen. New flow:

txflow-new-2

kloaec:
Great changes, thanks for that.

For consistency, I would recommend that you also use the same rewording on other flows, such as the "register descriptor" flow, which uses the "approve" term instead of "continue", for intermediary steps.

Another flow that I use every day: spending from Liana.
First step is "spend from known wallet"... "approve". This approve should be a "continue?" or even "review details". (because "continue" spending isn't really clear either!)

The rest of the PR is great, let's remove the "approve" where they are just a "next"/"continue" button.

bigspider:
> For consistency, I would recommend that you also use the same rewording on other flows, such as the "register descriptor" flow, which uses the "approve" term instead of "continue", for intermediary steps.

Yes, that's a good idea. I prefer to keep it for a future PR because I'm hoping to explore _not_ having a 'Continue' step at all: you just keep going right until the end, and then you have a single "Approve" step at the end.
In principle, that could be done for other flows as well, but it might be quite tricky because of how the current UX framework works.
I'll open a separate issue for the wallet registration flow.

> Another flow that I use every day: spending from Liana. First step is "spend from known wallet"... "approve". This approve should be a "continue?" or even "review details". (because "continue" spending isn't really clear either!)

It's already changed with 'Continue' in this PR, for example:
txflow-from-wallet

Rose2161 (approved):

Rose2161 (approved):

PR #259, opened 2024-05-06
Allow up to 5 data pushes in OP_RETURN outputs

bigspider:


src/common/script.c
```diff
@@ -130,66 +130,58 @@ int format_opscript_script(const uint8_t script[],
...
+            if (offset + hex_length > script_len) return -1;  // out of bounds
         } else if (opcode == OP_PUSHDATA1) {
-            // OP_RETURN OP_PUSHDATA1 <len:1-byte> <data:len bytes>
-            hex_offset += 2;
-            hex_length = script[2];
-
-            if (script_len != 1 + 1 + 1 + hex_length || hex_length > 80) return -1;
+            if (offset >= script_len) return -1;  // out of bounds for length byte
+            hex_length = script[offset++];
+            if (hex_length > 80 || offset + hex_length > script_len) return -1;
```
tdejoigny-ledger, line 158: I prefer when there are "{ }" when a "if" is used (easier to read and to maintain)

src/common/script.c
```diff
@@ -130,66 +130,58 @@ int format_opscript_script(const uint8_t script[],
...
 
         if (hex_length > 0) {
-            const char hex[] = "0123456789abcdef";
-
             out[out_ctr++] = '0';
             out[out_ctr++] = 'x';
             for (unsigned int i = 0; i < hex_length; i++) {
-                uint8_t data = script[hex_offset + i];
+                uint8_t data = script[offset + i];
                 out[out_ctr++] = hex[data / 16];
```
tdejoigny-ledger, line 175: should we use a bit shift here ?

tdejoigny-ledger (approved):
should be good to add some defines for relevant constants. It will be easier to understand

src/common/script.c
bigspider: Since it's paired with the `% 16` the next line, I think it's clearer with the division. The alternative would be to use `>> 4` in this line and `& 0x0F` in the next line, but I prefer the the division/modulus as it works for any base and it's more symmetric :)
The compiler would most likely optimize it (not that it matters, here).

src/common/script.c
bigspider: Changed with explicit brackets in 5fe7f05

tdejoigny-ledger (approved):

PR #261, opened 2024-06-18
Performance test suite

bigspider:
Adds a test suite to measure the performance of the app on a series of synthetically generated testcases.

The `Makefile` adds a new `AUTOAPPROVE_FOR_PERF_TESTS` that compiles a version of the app that automatically approves any request, in order to remove the UX interactions from the measurements. As a sanity precaution, compilation is aborted if this flag is enabled when compiling on mainnet.

The test suite can be ran either on speculos or on a real device; measured times on speculos are of course not a good indicator of the performance on real devices; nevertheless, the test suite is added as a CI job, since comparing the performance reports before and after a change can give a useful indicator about expected or unexpected changes in the running time. The performance report is also saved as a json artifact.

The PR adds an initial set of benchmarks related to signing different kinds of transactions. More extensive benchmarks can be added in the future.

Some legacy tests using the `txmaker` tool are deleted; no longer worth the maintenance burden.


yogh333 (dismissed):
LGTM

PR #262, opened 2024-06-26
Remove Nano S target

bigspider:
Removes the Nano S compilation target. Also removes some RAM and code size optimizations that are not longer worth the added code complexity.

The latest Nano S version is kept in the [nanos](https://github.com/LedgerHQ/app-bitcoin-new/tree/nanos) branch.


Makefile
```diff
@@ -113,9 +113,14 @@ else
     endif
 endif
 
+ifneq (,$(filter-out clean,$(MAKECMDGOALS)))
+  ifeq ($(TARGET_NAME),TARGET_NANOS)
+    $(error This repository is not compatible with the Nano S. Look for the 'nanos' branch for the latest code for Nano S.)
```
fbeutin-ledger, line 118:
```suggestion
    $(error This repository is not compatible with the Nano S. Look for the 'nanos' branch for the latest code for Nano S.)
    test
```

fbeutin-ledger (approved):

PR #272, opened 2024-07-17
Sync master branch to 2.2.4

bigspider:


.github/workflows/codeql-workflow.yml
```diff
@@ -17,8 +17,6 @@ jobs:
     strategy: 
       matrix:
         include:
-          - SDK: "$NANOS_SDK"
```
b0l0k, line 20: Is it on purpose?

bigspider: Yes, Nano S won't have future versions and the main branches are no longer compatible with Nano S starting from this version.

The last Nano S version is kept on its own `nanos` branch.

src/handler/register_wallet.c
```diff
@@ -123,44 +130,39 @@ void handler_register_wallet(dispatcher_context_t *dc, uint8_t protocol_version)
...
+                                                        cosigner_index,
+                                                        (uint8_t *) keys_info[cosigner_index],
+                                                        MAX_POLICY_KEY_INFO_LEN);
 
-        if (pubkey_info_len < 0) {
+        if (key_info_len < 0) {
             SEND_SW(dc, SW_INCORRECT_DATA);
-            ui_post_processing_confirm_wallet_registration(dc, false);
             return;
         }
```
b0l0k, line 156: I'd add a control on `key_info_len > MAX_POLICY_KEY_INFO_LEN + 1` even though I assume that `call_get_merkle_leaf_element` has already this control.

b0l0k (approved):
LGTM.
Happy to see that dropping the support of Nano S get rid of many memory trick.

PR #294, opened 2024-10-09
MuSig2 support

### LINKED ISSUE #177: Allow the client to only sign for a subset of key placeholders
Currently, the app only uses the `PSBT_IN_BIP32_DERIVATION`/`PSBT_IN_TAP_BIP32_DERIVATION` to figure out the right `change`/`address_index` pair for each input. However, there is no way of explicitly instructing _not_ to sign for some placeholders; this is a common need for miniscript-based wallets like Liana, where multiple spending paths might be controlled by the same device (with different derivations paths).

Rather than modifying the protocol, the simplest way might be to _not_ sign with a placeholder if the corresponding PSBT_GLOBAL_XPUB is missing in the PSBT.

### LINKED ISSUE #208: MuSig2 support
Extend wallet policies with `musig` support, as tentatively specified [here](https://github.com/achow101/bips/blob/musig2-descriptors/bip-musig2-descriptors.mediawiki).

For wallet policies, expressions of the form:

```
musig(KEY, ..., KEY)/<M,N>/*
musig(KEY, ..., KEY)/**
```
will suffice; `KEY` can be restricted to xpubs (although it's possible to generalize to other key expressions if there are use cases). It is unclear if there is any benefit in supporting things like `musig(KEY/**, ..., KEY/**)`, which seems [inherently inefficient](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2023-November/022132.html).

References
  - [BIP-0327](https://github.com/bitcoin/bips/blob/master/bip-0327.mediawiki)
  - The drafts BIPs for Musig2: https://github.com/bitcoin/bips/compare/master...achow101:bips:musig2.
  - PR for MuSig2 support in libsecp256k1: https://github.com/bitcoin-core/secp256k1/issues/1452

Tasks:
   - Bitcoin app
     - [x] Wallet policy parsing of `musig` descriptors
     - [x] Update specs for PSBT fields (aggregate key is now a plain key and not x-only)
     - [x] Aggregate xpub derivation for `musig` descriptor
     - [x] Address generation for policies with `musig` descriptor.
     - [x] Update to sorted `musig()` (per [this](https://github.com/bitcoin/bips/pull/1540#issuecomment-2059110591))
     - [x] Review sanity rules (e.g.: repeated xpubs in wallet policy) in the presence of `musig`
     - [x] Transaction-level signing session logic
     - [x] Pubnonce generation
     - [x] Partial signature in signing logic.
     - [x] Compute the aggregate key only once for the placeholder
     - [x] Sign the second round silently if both rounds are executed back-to-back.
   - Client libraries:
     - [x] Extend processing of YIELD to understand pubnonces and partial signatures (all clients)
     - [x] Add musig() xpub generation to python client
     - [x] Update to sorted `musig()` (per [this](https://github.com/bitcoin/bips/pull/1540#issuecomment-2059110591))
     - [x] <del>Add musig() xpub generation to Rust client</del> moved to #303
     - [x] <del>Add musig() xpub generation to JS client </del> moved to #303
   - UX
     - [x] Figure out what to show when no signature is produced, and only pubnonces (instead of Transaction Signed)
   - Tests:
     - [x] Address generation
     - [x] Implement cosigner logic and psbt partial signature aggregation in the python test suite
     - [x] <del>Extensive tests</del> moved to #302
     - [x] Figure out how to optimize the CI to not recompile bitcoin every time

bigspider:
Add full support for `musig()` in wallet policies.

Unlike typical signatures, where the client sends a transaction to the device and receives a signed transaction back, MuSig2  is a 2-party protocol among a group of cosigners, where:
- during the first round, all the cosigners generate a secret nonce, and share with each other a _pubnonce_
- Once all the pubnonces are available, the cosigners can proceed to aggregate the _pubnonces_, producing the aggregate _aggnonce_, and produce a _musig partial signature_

Any (untrusted) coordinator, once all the partial signatures are collected, can produce the final signature.

### Security considerations

Special care needs to be taken in the handling of nonces, and the way nonces are persisted in storage.
See more in the `musig.md` document added as part of the PR.

### UX consideration

As it is a two round protocol, there were some choices to be made on how to handle it: software wallets will in fact send the same PSBT to the device _twice_ in order to complete the protocol.

In order to minimize the UX impact for the device - but also for wallets implementing it, the following choice was made:
- _the first round is executed silently_ - that is without requiring any user confirmation

This is not a security problem, because the first round of MuSig2 just produces and stores a random number; private keys are not even accessed.

In cases when the other cosigners are online at the same time, this allows software wallets to use MuSig2 while essentially providing identical UX to a non-MuSig2 wallet (connect the device, <first round happens silently>, inspect and confirm the transaction).

If the first and second round happen in two separate sessions, only the second one will require user confirmation, while of course the first one still needs the device to be unlocked and connected with the bitcoin app open.

### BIPs related to this PR

The details of the signing protocol are detailed in [BIP-327](https://github.com/bitcoin/bips/blob/master/bip-0327.mediawiki), but there are other related BIPs: [BIP-328](https://github.com/bitcoin/bips/blob/master/bip-0328.mediawiki), [BIP-390](https://github.com/bitcoin/bips/blob/master/bip-0390.mediawiki), [BIP-373](https://github.com/bitcoin/bips/blob/master/bip-0373.mediawiki).
[BIP-388](https://github.com/bitcoin/bips/blob/master/bip-0388.mediawiki) is [in process of being updated](https://github.com/bitcoin/bips/pull/1697).

### Structure of the implementation

The PR required some refactoring that have a fairly big impact on the codebase, particularly in order to generalize the 'key expressions' in wallet policies to support the `musig()` expression, whereas in the past it was always limited to a single xpub; the code to handle the PSBT fields in the transaction and identify the correct derivation path for each input also needed to be generalized.
However, most of the musig-related code is relatively self-contained and modularized in the following files:

- `doc/musig.md` - Technical details, and security details, about the MuSig2 feature.
- `src/handler/sign_psbt/musig_signing.{c,j}` - portions of the `sign_psbt` handler that are handling the logic of Round 1 and Round 2 of MuSig2,
- `src/musig/musig.{c,h}` - implementation of the relevant algorithms defined in [BIP-327](https://github.com/bitcoin/bips/blob/master/bip-0327.mediawiki), closely matching the [Python reference implementation](https://github.com/bitcoin/bips/blob/master/bip-0327/reference.py)
- `src/musig/musig_sessions.{c,h}` - encapsulates the logic to handle the persistence in NVRAM after the end of MuSig's round 1, until the beginning of Round 2.

### Status of test coverage

There are some e2e tests with the corresponding bitcoin-core implementation which is still a WIP and does not support complex policies. Testing will be needed with musig nested inside miniscript as soon as possible - to be done in a separate PR ahead of release. Created issue #302 for tracking.

### Incidental changes

Also improves the signing flow to only sign for spending paths that are filled in in the PSBT (that is, the corresponding `PSBT_IN_BIP32_DERIVATION` or  `PSBT_IN_TAP_BIP32_DERIVATION` fields are present), optimizing certain use cases for wallet policies with multiple alternative spending paths.

Closes: #208 
Closes: #177


bitcoin_client/ledger_bitcoin/bip0327.py
```diff
@@ -0,0 +1,177 @@
...
+# Assertions are used to (1) satisfy the type-checking system, and (2) check for
+# inconvenient events that can't happen except with negligible probability (e.g.
+# output of a hash function is 0) and can't be manually triggered by any
+# signer.
+
+# This exception is raised if a party (signer or nonce aggregator) sends invalid
+# values. Actual implementations should not crash when receiving invalid
+# contributions. Instead, they should hold the offending party accountable.
+class InvalidContributionError(Exception):
+    def __init__(self, signer, contrib):
```
lpascal-ledger, line 108:
```suggestion
    def __init__(self, signer: Optional[int], contrib: str):
```

bitcoin_client/ledger_bitcoin/wallet.py
```diff
@@ -58,34 +59,52 @@ def n_keys(self) -> int:
...
             write_varint(len(self.descriptor_template.encode())),
-            self.descriptor_template.encode() if self.version == WalletType.WALLET_POLICY_V1 else descriptor_template_sha256,
+            self.descriptor_template.encode(
+            ) if self.version == WalletType.WALLET_POLICY_V1 else descriptor_template_sha256,
             write_varint(len(self.keys_info)),
             MerkleTree(keys_info_hashes).root
         ])
 
-    def get_descriptor(self, change: bool) -> str:
+    def get_descriptor(self, change: Union[bool, None]) -> str:
```
lpascal-ledger, line 77: I see you use `Optional[type]` elsewhere, is this technically equivalent syntax here to express a subtlety compared to `Optional`?

doc/bitcoin.md
```diff
@@ -229,17 +229,33 @@ No output data; the signature are returned using the YIELD client command.
 
 Using the information in the PSBT and the wallet description, this command verifies what inputs are internal and what outputs match the pattern for a change address. After validating all the external outputs and the transaction fee with the user, it signs each of the internal inputs; each signature is sent to the client using the YIELD command, in the format described below. If multiple key placeholders of the wallet policy are internal, the process is repeated for each of them.
 
+For a registered wallet, the hmac must be correct.
+
+For a default wallet, `hmac` must be equal to 32 bytes `0`.
+
+The data yielded has with a `<tag_or_input_index>`, represented as a bitcoin-style varint; the rest of the message depends on the value of `<tag_or_input_index>`, as specified in the following subsections.
```
lpascal-ledger: I'm no native English speaker but "The data yielded has with [...]" seems strange to me and I'm not sure how to understand it.

doc/musig.md
```diff
@@ -0,0 +1,87 @@
...
+  - Verify that the pubnonce contained in the PSBT matches the one synthetically recomputed.
+  - Continue the signing flow as per BIP-0327, generating the partial signature.
+
+### Security considerations
+#### State reuse avoidance
+Storing the session in persistent memory only at the end of Phase 1, and deleting it before beginning Phase 2 simplifies auditing and making sure that there is no reuse of state across signing sessions.
+
+#### Security of synthetic randomness
+
+Generating $rand_{i, j}$ synthetically is not a problem, since the $rand\_root$ value is kept secret and never leaves the device. This ensures that all the values produced for different $i$ and $j$ not predictable for an attacker.
```
lpascal-ledger: ### 
```suggestion
Generating $rand_{i, j}$ synthetically is not a problem, since the $rand\_root$ value is kept secret and never leaves the device. This ensures that all the values produced for different $i$ and $j$ are not predictable for an attacker.
```

doc/musig.md
```diff
@@ -0,0 +1,87 @@
...
+
+#### Malleability of the PSBT
+If the optional parameters are passed to the _NonceGen_ function, they will depend on the transaction data present in the PSBT. Therefore, there is no guarantee that they will be unchanged the next time the PSBT is provided.
+
+However, that does not constitute a security risk, as those parameters are only used as additional sources of entropy in _NonceGen_. A malicious software wallet can't affect the _secnonce_/_pubnonce_ pairs in any predictable way. Changing any of the parameters used in _NonceGen_ would cause a failure during Phase 2, as the recomputed _pubnonce_ would not match the one in the psbt.
+
+### Generalization to multiple PSBT signing sessions
+
+The approach described above assumes that no attempt to sign a PSBT containing for a wallet policy containing `musig()` keys is initiated while a session is already in progress.
```
lpascal-ledger:
```suggestion
The approach described above assumes that no attempt to sign a PSBT for a wallet policy containing `musig()` keys is initiated while a session is already in progress.
```

src/common/wallet.c
```diff
@@ -424,23 +424,120 @@ int parse_policy_map_key_info(buffer_t *buffer, policy_map_key_info_t *out, int
...
+                break;
+            } else if (c != ',') {
+                return WITH_ERROR(-1, "Invalid character in musig; expected ',' or ')'");
+            }
+        }
+
+        if (n_musig_keys < 2) {
+            return WITH_ERROR(-1, "musig must have at least 2 key indexes");
+        }
+        if (n_musig_keys > MAX_PUBKEYS_PER_MUSIG) {
```
lpascal-ledger: Shouldn't it be checked inside the `while` loop rather than after? If the number of keys is arbitrary, can't it be damaging for your stack (and/or be slow to be computed, just to raise an error after)?

lpascal-ledger (commented):
Reviewed the C part mostly, need to dive more into the Python bits.

bitcoin_client/ledger_bitcoin/bip0327.py
bigspider: As this file is taken from a [reference implementation](https://github.com/bitcoin/bips/blob/master/bip-0327/reference.py) in the BIP, I prefer not to make any changes.

bitcoin_client/ledger_bitcoin/wallet.py
bigspider: Yeah, it's semantically equivalent; I preferred this version because the behavior with `None` is completely distinct from the `True`/`False` ones, as opposed to a situation where a parameter can be omitted (e.g. because there's a default value).

doc/bitcoin.md
bigspider: Fixed in 64897008ab6643c1cb698fcc6b9abb771f9e4540.

doc/musig.md
bigspider: Fixed in 64897008ab6643c1cb698fcc6b9abb771f9e4540

doc/musig.md
bigspider: This part was actually incomplete, as it doesn't say how the actual implementation works (just claiming it can be generalized). Added a short description in 64897008ab6643c1cb698fcc6b9abb771f9e4540.

src/common/wallet.c
bigspider: Good point, indeed this check is technically redundant as the loop above guarantees that `n_musig_keys <= MAX_PUBKEYS_PER_MUSIG` on exit.

Probably worth keeping the the check outside the loop as a `LEDGER_ASSERT`, which most likely the compiler is smart enough to optimize away.

Done in 64897008ab6643c1cb698fcc6b9abb771f9e4540.

bitcoin_client/ledger_bitcoin/wallet.py
lpascal-ledger: Yeah fair point vs. a `Optional[bool] = None` kwarg.

doc/musig.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Generalization to multiple PSBT signing sessions
+
+The approach described above assumes that no attempt to sign a PSBT for a wallet policy containing `musig()` keys is initiated while a session is already in progress.
+
+In order to generalize this to an arbitrary number of parallel signing sessions, one can identify each signing session with a `psbt_session_id`. Such `psbt_session_id` should deterministically depend on the transaction being signed (ignoring all the other PSBT fields), and the wallet policy being signed. In praticular, the computed `psbt_session_id` should be identical between Round 1 and Round 2 of the protocol. Note that malicious collisions of the `psbt_session_id` (for example by tampering with some details of the PSBT, like the SIGHASH flags) _are_ possible, but they do not constitute a security risk.
```
lpascal-ledger, line 87:
```suggestion
In order to generalize this to an arbitrary number of parallel signing sessions, one can identify each signing session with a `psbt_session_id`. Such `psbt_session_id` should deterministically depend on the transaction being signed (ignoring all the other PSBT fields), and the wallet policy being signed. In particular, the computed `psbt_session_id` should be identical between Round 1 and Round 2 of the protocol. Note that malicious collisions of the `psbt_session_id` (for example by tampering with some details of the PSBT, like the SIGHASH flags) _are_ possible, but they do not constitute a security risk.
```

May I suggest you to add a `codespell-project/actions-codespell` action in your workflows (example in [Speculos](https://github.com/LedgerHQ/speculos/blob/master/.github/workflows/fast-checks.yml#L60-L74))?
It can be annoying sometimes because it can raise false positives and is a bit annoying to configure, but still it generally is very useful.

test_utils/musig2.py
```diff
@@ -0,0 +1,862 @@
...
+    This is minimal implementation in order to enable iterating over the placeholders,
+    and compile the corresponding leaf scripts.
+    """
+
+    def __init__(self, key: KeyPlaceholder, tree=Optional[Tree]):
+        self.key: KeyPlaceholder = key
+        self.tree: Optional[Tree] = tree
+
+    @classmethod
+    def from_string(cls, input_string):
```
lpascal-ledger:
```suggestion
    def from_string(cls, input_string: str) -> "TrDescriptorTemplate":
```

test_utils/musig2.py
```diff
@@ -0,0 +1,862 @@
...
+        parser = cls.Parser(input_string.replace("/**", "/<0;1>/*"))
+        return parser.parse()
+
+    class Parser:
+        def __init__(self, input):
+            self.input = input
+            self.index = 0
+            self.length = len(input)
+
+        def parse(self):
```
lpascal-ledger:
```suggestion
        def parse(self) -> "TrDescriptorTemplate":
```

test_utils/musig2.py
```diff
@@ -0,0 +1,862 @@
...
+        self.key: KeyPlaceholder = key
+        self.tree: Optional[Tree] = tree
+
+    @classmethod
+    def from_string(cls, input_string):
+        parser = cls.Parser(input_string.replace("/**", "/<0;1>/*"))
+        return parser.parse()
+
+    class Parser:
+        def __init__(self, input):
```
lpascal-ledger, line 296:
```suggestion
        def __init__(self, input: str):
```

test_utils/musig2.py
```diff
@@ -0,0 +1,862 @@
...
+                if self.peek() == ',':
+                    self.consume(',')
+                    tree = self.parse_tree()
+                self.consume(')')
+                return TrDescriptorTemplate(key, tree)
+            else:
+                raise Exception(
+                    "Syntax error: Input does not start with 'tr('")
+
+        def parse_keyplaceholder(self):
```
lpascal-ledger:
```suggestion
        def parse_keyplaceholder(self) -> KeyPlaceholder:
```

test_utils/musig2.py
```diff
@@ -0,0 +1,862 @@
...
+            while self.index < self.length and (nesting > 0 or self.input[self.index] not in ('}', ',', ')')):
+                if self.input[self.index] == '(':
+                    nesting = nesting + 1
+                elif self.input[self.index] == ')':
+                    nesting = nesting - 1
+
+                self.index += 1
+            return self.input[start:self.index]
+
+        def parse_key_indexes(self):
```
lpascal-ledger:
```suggestion
        def parse_key_indexes(self) -> List[int]:
```

test_utils/musig2.py
```diff
@@ -0,0 +1,862 @@
...
+            nums = []
+            self.consume('@')
+            nums.append(self.parse_num())
+            while self.peek() == ',':
+                self.consume(',@')
+                nums.append(self.parse_num())
+            return nums
+
+        def parse_num(self):
```
lpascal-ledger:
```suggestion
        def parse_num(self) -> int:
```

test_utils/musig2.py
```diff
@@ -0,0 +1,862 @@
...
+            start = self.index
+            while self.index < self.length and self.input[self.index].isdigit():
+                self.index += 1
+            return int(self.input[start:self.index])
+
+        def consume(self, char):
```
lpascal-ledger:
```suggestion
        def consume(self, char: str) -> None:
```

tests/conftest.py
```diff
@@ -1,32 +1,32 @@
...
-from test_utils import segwit_addr
-
-import shutil
-import subprocess
-from time import sleep
-from decimal import Decimal
-from pathlib import Path
+absolute_path = os.path.dirname(os.path.abspath(__file__))  # noqa: E402
+relative_bitcoin_path = ('../bitcoin_client')  # noqa: E402
+absolute_bitcoin_client_path = os.path.join(
```
lpascal-ledger, line 7: This variable does not seem to be used anywhere (as well as the same one in `tests_perf/conftests`).

tests/conftest.py
```diff
@@ -1,32 +1,32 @@
...
+    os.path.dirname(os.path.abspath(__file__)), '../')   # noqa: E402
+sys.path.append(os.path.join(absolute_path, relative_bitcoin_path))  # noqa: E402
```
lpascal-ledger, line 9: You pobably already know `pathlib.Path` and have your reason not to use it here, still I think it is interesting to consider it on cluttered lines like these.

The direct equivalent would be
```python
from pathlib import Path
absolute_path = Path(__file__).absolute().parent
relative_bitcoin_path = Path("../bitcoin_client")
absolute_bitcoin_client_path = Path(__file__).absolute().parent.parent
sys.path.append(str(absolute_path / relative_bitcoin_path)) 
```

but given the unused variable, it rather is:

```suggestion
from pathlib import Path
absolute_path = Path(__file__).absolute().parent  # noqa: E402
sys.path.append(str(absolute_path / "bitcoin_client"))  # noqa: E402
```

It also now works with (most) builtin libraries, such as shutil:

```python
    shutil.copy(Path(__file__).parent / "bitcoin.conf"), BITCOIN_DIRNAME)
    # or in your case
    shutil.copy(absolute_path / "bitcoin.conf", BITCOIN_DIRNAME)
```

lpascal-ledger (dismissed):
Nothing interesting to say, that all sounds fair :+1:

doc/musig.md
bigspider: Created issue #306, good idea!

tests/conftest.py
bigspider: Indeed, removed in 540158f46590ffb52520285fe3a2731501d82798

tests/conftest.py
bigspider: Thanks! Done in 540158f46590ffb52520285fe3a2731501d82798

bigspider: Reverted, somehow it broke the CI. Leaving it for a future PR.

lpascal-ledger (approved):

PR #318, opened 2025-02-28
Increase `MAX_N_KEYS_IN_WALLET_POLICY` to 15

### LINKED ISSUE #315: Increase the maximum number of key expressions in the policy
Version 2.2.4 introduced a technical maximum limit of 10 keys in a wallet policy. However, it has been requested to increase this limit to at least 15 in order to allow managing keys for complex federations.

### LINKED ISSUE #317: 0x6a82 Error in BitcoinInsType.REGISTER_WALLET
Trying to sign a Bitcoin multisig transaction in Electrum using a Nano Flex, Nano X or Nano S Plus with Bitcoin app version 2.3.0, throws:

('0x6a82', 'Error in <BitcoinInsType.REGISTER_WALLET: 2> command','')

Maybe to mention, we use a native segwit multisig p2wsh, signer threshold is 4 of 15, with a non-standard derivation path on the last level m/48h/0h/0h/**1h** instead of m/48h/0h/0h/**2h** which would be default for native segwit multisig.

The same transaction could be signed without any issues from a Ledger Nano S using the app in version 2.2.5 and 2.2.3 and potentially older versions in the past as well, but 2.3.0 seems to fail here.

bigspider:
This allows complex wallet policies with a larger number of cosigners.
This was supported in past versions, but the cap to 10 keys was introduced in version 2.2.4 together with the UX refactoring.

The memory occupation increases by 160 bytes for devices using NBGL, due to reserving a bigger number of UX pairs (stored in a global).

There is otherwise no increase in memory usage in SIGN_PSBT, and the increase in REGISTER_WALLET is not problematic - plenty of memory is available there.

b8b7d0ccc fixes the corresponding test, and adds it to the CI, but only on merges and not for PRs, as it's rather slow.

Closes: #315
Closes: #317

### DIFF
--- a/src/ui/display_nbgl.c
+++ b/src/ui/display_nbgl.c
@@ -15,7 +15,7 @@ static const char *confirmed_status;  // text displayed in confirmation page (af
 static const char *rejected_status;   // text displayed in rejection page (after reject confirmed)
 static bool show_message_start_page;
 
-#define N_UX_PAIRS 13
+#define N_UX_PAIRS 18
 
 static nbgl_layoutTagValue_t pairs[N_UX_PAIRS];
 static nbgl_layoutTagValueList_t pairList;
--- a/test_utils/fixtures.py
+++ b/test_utils/fixtures.py
@@ -52,13 +52,13 @@ def pytest_addoption(parser):
     parser.addoption("--hid", action="store_true")
     parser.addoption("--headless", action="store_true")
     parser.addoption(
-        "--enableslowtests",
+        "--enable_slow_tests",
         action="store",
         nargs="?",
         const=1,  # Default to 1 if the flag is provided without a value
         default=0,  # Default to 0 if the flag is not provided
         type=int,
-        help="Enable slow tests: 0 (default), 1 if flag is present, or specify an integer",
+        help="If 0, skip slow tests. If set or equal to 1, enable some slow tests. Greater values enable even slower tests.",
     )
     parser.addoption("--model", action="store", default="nanosp")
 
@@ -99,7 +99,7 @@ def headless(pytestconfig) -> bool:
 
 @pytest.fixture
 def enable_slow_tests(pytestconfig) -> int:
-    return pytestconfig.getoption("enableslowtests")
+    return pytestconfig.getoption("enable_slow_tests")
 
 
 @pytest.fixture
--- a/test_utils/fixtures.py
+++ b/test_utils/fixtures.py
@@ -51,12 +51,20 @@ def get_app_version() -> str:
 def pytest_addoption(parser):
     parser.addoption("--hid", action="store_true")
     parser.addoption("--headless", action="store_true")
-    parser.addoption("--enableslowtests", action="store_true")
+    parser.addoption(
+        "--enableslowtests",
+        action="store",
+        nargs="?",
+        const=1,  # Default to 1 if the flag is provided without a value
+        default=0,  # Default to 0 if the flag is not provided
+        type=int,
+        help="Enable slow tests: 0 (default), 1 if flag is present, or specify an integer",
+    )
     parser.addoption("--model", action="store", default="nanosp")
 
 
 @pytest.fixture(scope="module")
-def sw_h_path():
+def sw_h_path() -> Path:
     # sw.h should be in src/boilerplate/sw.h
     sw_h_path = repo_root_path / "src" / "boilerplate" / "sw.h"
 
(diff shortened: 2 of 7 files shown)


test_utils/fixtures.py
```diff
@@ -51,12 +51,20 @@ def get_app_version() -> str:
 def pytest_addoption(parser):
     parser.addoption("--hid", action="store_true")
     parser.addoption("--headless", action="store_true")
-    parser.addoption("--enableslowtests", action="store_true")
+    parser.addoption(
+        "--enableslowtests",
```
iartemov-ledger: https://peps.python.org/pep-0008/#function-and-variable-names
I did not find a specific convention for input arguments, but it falls back finally to the variables' one - so snake style, "with words separated by underscores as necessary to improve readability."

tdejoigny-ledger (dismissed):

tests/test_e2e_multisig.py
```diff
@@ -198,17 +198,17 @@ def test_e2e_multisig_multiple_internal_keys(navigator: Navigator, firmware: Fir
...
 @pytest.mark.timeout(0)  # disable timeout
-def test_e2e_multisig_16_of_16(navigator: Navigator, firmware: Firmware, client: RaggerClient, test_name: str, rpc: AuthServiceProxy, rpc_test_wallet, speculos_globals: SpeculosGlobals, enable_slow_tests: bool):
-    # Largest supported multisig with sortedmulti.
+def test_e2e_multisig_15keys(navigator: Navigator, firmware: Firmware, client: RaggerClient, test_name: str, rpc: AuthServiceProxy, rpc_test_wallet, speculos_globals: SpeculosGlobals, enable_slow_tests: int):
+    # Largest supported quorum in a multisig.
     # The time for an end-to-end execution on a real Ledger Nano S (including user's input) is about 520 seconds.
 
-    # slow test, disabled by default
+    # slow test, only run it if --enableslowtests is set
     if not enable_slow_tests:
```
iartemov-ledger: Maybe `if enable_slow_tests == 0' ?
To reflect the fact that it is integer now and to align with   the check below in `tests/test_sign_psbt.py` ?

src/ui/display_nbgl.c
```diff
@@ -15,7 +15,7 @@ static const char *confirmed_status;  // text displayed in confirmation page (af
 static const char *rejected_status;   // text displayed in rejection page (after reject confirmed)
 static bool show_message_start_page;
 
-#define N_UX_PAIRS 13
+#define N_UX_PAIRS 18
```
iartemov-ledger, line 18: Would it be possible to derive the value from MAX_N_KEYS_IN_WALLET_POLICY ?
As on line 324 ?

tests/test_e2e_multisig.py
```diff
@@ -198,17 +198,17 @@ def test_e2e_multisig_multiple_internal_keys(navigator: Navigator, firmware: Fir
...
         pytest.skip()
 
     core_wallet_names: List[str] = []
     core_xpub_origs: List[str] = []
-    for _ in range(15):
+    for _ in range(14):
```
iartemov-ledger, line 211: It is done for [0..13] so 14 times. Is it expected, why not 15 ?

test_utils/fixtures.py
```diff
@@ -51,12 +51,20 @@ def get_app_version() -> str:
...
+        action="store",
+        nargs="?",
+        const=1,  # Default to 1 if the flag is provided without a value
+        default=0,  # Default to 0 if the flag is not provided
+        type=int,
+        help="Enable slow tests: 0 (default), 1 if flag is present, or specify an integer",
```
iartemov-ledger: Maybe to add that at `0` the "enable-slow-tests" are not run and the other values greater than zero are the "levels" (?)  of test running ? 

Or maybe even a choice value - `choices=['no_run', 'no_sign_psbt', 'all']` ?

tests/test_e2e_multisig.py
bigspider: Yes, because one of the cosigners is the device, all the others are wallets managed by bitcoin-core.

src/ui/display_nbgl.c
bigspider: the problem is that the various flow in the file add different constraints, so you'd need to take the MAX of all the values you compute from each of the constraints. That's why I added the all the static asserts instead.

Not sure there is an easy way, but I'd leave it for a separate PR!

test_utils/fixtures.py
bigspider: Added_dashes_in 42b9722e3091c374380b5871423adb9a3a9fd7bb

tests/test_e2e_multisig.py
bigspider: Changed to `< 1` for symmetry with `test_sign_psbt_singlesig_wpkh_512to256` where it's `< 2`
42b9722e3091c374380b5871423adb9a3a9fd7bb

test_utils/fixtures.py
bigspider: Tried to make the help text more explicit in 42b9722e3091c374380b5871423adb9a3a9fd7bb.

src/ui/display_nbgl.c
iartemov-ledger: It would make sense so far to add a comment about it for `N_UX_PAIRS` define ;)

test_utils/fixtures.py
iartemov-ledger: It is more clear!

test_utils/fixtures.py
iartemov-ledger: It is really more readable , thank you!
You say dashes and indeed the convention for the arguments might be that it uses dashes and then it is automatically converted to underscores, but only for optional arguments: https://docs.python.org/dev/library/argparse.html#dest.
But I'm not sure for this argument in particular and anyway it is a detail and we have already won in readability !

iartemov-ledger (approved):

PR #320, opened 2025-03-06
Prepare release 2.4.0

bigspider:
Add MuSig2 test with 5 keys in both the keypath and the script path
That's the maximum currently supported, it's good to have a test for it.

Prepares release 2.4.0.


src/ui/menu_bagl.c
```diff
@@ -57,7 +57,7 @@ UX_FLOW(ux_menu_main_flow_bitcoin_testnet,
         &ux_menu_exit_step,
         FLOW_LOOP);
 
-UX_STEP_NOCB(ux_menu_info_step, bn, {"Bitcoin App", "(c) 2024 Ledger"});
+UX_STEP_NOCB(ux_menu_info_step, bn, {"Bitcoin App", "(c) 2025 Ledger"});
```
iartemov-ledger, line 60: Makes sense to update the copyright header as well ?

src/ui/menu_nbgl.c
```diff
@@ -23,7 +23,7 @@
 
 #define SETTING_INFO_NB 3
 static const char* const INFO_TYPES[SETTING_INFO_NB] = {"Version", "Developer", "Copyright"};
-static const char* const INFO_CONTENTS[SETTING_INFO_NB] = {APPVERSION, "Ledger", "(c) 2024 Ledger"};
+static const char* const INFO_CONTENTS[SETTING_INFO_NB] = {APPVERSION, "Ledger", "(c) 2025 Ledger"};
```
iartemov-ledger, line 26: Makes sense to update the copyright header as well ?

tests/test_e2e_musig2.py
```diff
@@ -320,6 +320,46 @@ def test_e2e_musig2_3of3keypath_decaying_scriptpath(navigator: Navigator, firmwa
                         e2e_register_wallet_instruction(firmware, wallet_policy.n_keys), e2e_sign_psbt_instruction(firmware), test_name)
 
 
+def test_e2e_musig2_5of5(navigator: Navigator, firmware: Firmware, client: RaggerClient,
+                         test_name: str, rpc, rpc_test_wallet, speculos_globals: SpeculosGlobals, enable_slow_tests: int):
+    # 5 is the maximum number of keys we support have in a musig2 policy
+    # In this test, we have a musig with 5 keys in both the keypath and the script path
+
+    # slow test, only run it if --enable_slow_tests is set to a value greater than 0
+    if enable_slow_tests < 1:
```
iartemov-ledger, line 329: This test is skipped  in the context of the PR. 
It will be executed only on push to `develop` https://github.com/LedgerHQ/app-bitcoin-new/blob/musig2-moretests/.github/workflows/build_and_functional_tests.yml#L36, right ?
I wonder if it is not too late ?

tests/test_e2e_musig2.py
```diff
@@ -320,6 +320,46 @@ def test_e2e_musig2_3of3keypath_decaying_scriptpath(navigator: Navigator, firmwa
                         e2e_register_wallet_instruction(firmware, wallet_policy.n_keys), e2e_sign_psbt_instruction(firmware), test_name)
 
 
+def test_e2e_musig2_5of5(navigator: Navigator, firmware: Firmware, client: RaggerClient,
```
iartemov-ledger, line 323: Does this test use Speculos?  Doesn't it generate some snapshots ? 
I believe that not as they are absent, but just for a general understanding ;)

bitcoin_client/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 Dates are in `dd-mm-yyyy` format.
 
+## [0.4.0] - 06-03-2025
```
iartemov-ledger, line 10: Should you also update https://github.com/LedgerHQ/app-bitcoin-new/blob/develop/CHANGELOG.md ?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Dates are in `dd-mm-yyyy` format.
 
 - For wallet policies with multiple internal spending paths, the app will only sign for key expressions for which the corresponding `PSBT_IN_BIP32_DERIVATION` or `PSBT_IN_TAP_BIP32_DERIVATION` is present in the PSBT. This improves performance when signing for certain spending paths is not desired.
 - Increased the maximum supported number of cosigners in a wallet policy to 15.
+- Some common errors now return an [error code](src/error_codes.h) in addition to the usual status word.
```
iartemov-ledger, line 20: Updating "[2.4.0] - TBD" part would be the latest step ?

tests/test_e2e_musig2.py
bigspider: No, all the e2e tests don't generate screenshots. They all test a full signing flow (generating address, funding, signing the transaction and send it with bitcoin-core).

The screenshots are used instead on the tests for the individual command handlers.

tests/test_e2e_musig2.py
bigspider: The idea is to keep the running time shorter while PR is open, only doing some tests when PR's are merged to `develop` or `master`.
Some tests are quite slow and unlikely to fail (like the ones that are just stress-testing the limits of the app - like this test doing a Musig 5-of-5, or the one for multisig with 15 keys.

Of course, it could happen that some test only fails in these extreme cases - but it's probably going to be quite rare, and we'll still discover it before the app is released. In fact, I will probably move more of the slow tests in the `--enable_slow_tests` group in the future.

CHANGELOG.md
bigspider: Thanks, done in e90dd3ceea9f9c3685410721dcafe79521dfa06c.

src/ui/menu_nbgl.c
bigspider: Done in bd6a7365d5ea65c5f9070e49bdd17e5b62934aa1

src/ui/menu_bagl.c
bigspider: Done in bd6a7365d5ea65c5f9070e49bdd17e5b62934aa1

tests/test_e2e_musig2.py
iartemov-ledger: Ok. Thank you.

Makefile
```diff
@@ -1,6 +1,6 @@
 # ****************************************************************************
 #    Ledger App for Bitcoin
-#    (c) 2024 Ledger SAS.
+#    (c) 2025 Ledger SAS.
```
iartemov-ledger, line 3: Ah, I'm sorry to be picky. 
But usually the previous years are not overwritten,  just news ones are added at the end, like in this recommendaiton: https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html
It can be also a range as here https://github.com/LedgerHQ/app-ethereum/blob/develop/src/main.c#L3, it depends on convention. 
But it is important to keep the history IMHO. 
The remark applies to this and below copyright updates ;)

bigspider: As we never did it in this codebase and fixing it properly requires me to go through the history of each file, I'd say we can leave it for a separate PR.

iartemov-ledger: OK!

iartemov-ledger (approved):

PR #327, opened 2025-05-15
Update instructions about dependencies to install before running tests

bigspider:
Noticed while tying to run tests on a new machine.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -4,6 +4,12 @@ These tests are implemented in Python and can be executed either using the [Spec
 
 All the commands in this folder are meant to be ran from the `tests` folder, not from the root.
 
+Install the following system dependencies:
+
+```
+sudo apt-get install -y python3-pytest autoconf automake build-essential libffi-dev libtool pkg-config python3-dev
+```
+
 Python dependencies are listed in [requirements.txt](../requirements.txt), install them using [pip](https://pypi.org/project/pip/) from the root of the repository:
 
 ```


tests/README.md
```diff
@@ -4,6 +4,12 @@ These tests are implemented in Python and can be executed either using the [Spec
 
 All the commands in this folder are meant to be ran from the `tests` folder, not from the root.
 
+Install the following system dependencies:
+
+```
+sudo apt-get install -y python3-pytest autoconf automake build-essential libffi-dev libtool pkg-config python3-dev
```
iartemov-ledger, line 10: 2 little questions:
* could not we just point out to https://github.com/LedgerHQ/app-bitcoin-new/blob/develop/.github/workflows/Dockerfile either to be used for tests or even as documentation
* there seems to be a difference between the  list of packages suggested here and the ones installed in the docker. OK the initial Linux version may be different, should not we try to align/clean-up ?

bigspider: I don't run tests inside the Docker container on my machine, and I don't think tests should recommend that.

The list of packages is `pytest` plus all the packages required to build the python wheel for coincurve (taken from [here](https://github.com/ofek/coincurve/blob/master/docs/install.md#debianubuntu), which was failing to build on my machine.

iartemov-ledger: I'm not sure about it. 
It migth be better to have a "single point of failure" - one place that describes the environment (testing one in this case) - the one that is regularly tested by CI. The text/manual version is  always obsolete. 
Also, for example now  I'm trying to run and debug the JS-based tests and I would not like to drag Node JS and all what it takes with .

Anyway approving this minor change.

iartemov-ledger (approved):

PR #325, opened 2025-03-31
Nbgl migration for Nano devices

iartemov-ledger:
**In this PR:**
- [x]  Code base migration to NBGL for Nano devices
- [x]  Test adaptation 
- [x]  NanoX snapshots update (expected to be :green_heart:)
- [x]  NanoS+ snapshots update  (so tests failures are expected so far)  - will do later (after updated Product Guidance)  
- [x]  fix Tests on mainnet
- [x] fix Tests with the JS library  - will do later (after updated Product Guidance)  
- [x] remove `dispatcher_context_t *context` from `io_ui_process()` and all the callers  ? - will not do in this PR, the context may be useful later, for other needs 
- [x] App version update probably ?  The versions in `app-bitcoin-new` are usually bumped in separate PRs on demand
- [x]  Wait and address design update
- [x] to get, discuss and address feedback on https://ledgerhq.atlassian.net/wiki/spaces/FW/whiteboard/5704286287?atl_f=PAGETREE from stakeholders (eSW, Product, Design)  
     - in yellow-background comments - the difference summary
     - in red-background comments - the things that seem weird
- [x] Snapshots for long tests (`tests/test_sign_psbt.py::test_sign_psbt_singlesig_wpkh_512to256` fails with timeout (a several minutes one), even before present changes. Did it work recently ?)
    - Among SKIPPED tests non generates snapshots for Nano*:
        - tests/test_e2e_multisig.py::test_e2e_multisig_15keys
        - tests/test_e2e_musig2.py::test_e2e_musig2_5of5
        - tests/test_get_extended_pubkey.py::test_get_extended_pubkey_non_standard_reject_early
        - tests/test_sign_psbt.py::test_sign_psbt_singlesig_wpkh_512to256  
- [x] to take https://ledgerhq.atlassian.net/browse/B2CA-2132 into account
- [x] to update the wiki sequences: https://ledgerhq.atlassian.net/wiki/spaces/FW/whiteboard/5797675260?atl_f=PAGETREE
- [x] to clean-up the git history

**Later:**
- [ ] Refactoring to  `nbgl_useCaseAdvancedReview()` or `nbgl_useCaseReviewLight()`
- [ ] see if we can convert Stream uses case to normal one having introduced a memory limit

PR #328, opened 2025-05-23
Nbgl on nano fixes

iartemov-ledger:
Addressing https://ledgerhq.atlassian.net/browse/B2CA-2132?focusedCommentId=553816 namely:

- [x] icons redesign follow-up (Home scree: white background and the same as on Dash Board; Operation: black screen/inversed)
- [x]  More detailed "Loading transaction" and "Loading message" instead of generic "Processing" 
- [x]  Using SDK-provided `app_exit()`
- [x] Fixing the return to Home menu forgotten in https://github.com/LedgerHQ/app-bitcoin-new/pull/325
- [ ]  On transaction using LL Nano screen freeze for some time until finally the "Processing" appears (https://ledgerhq.atlassian.net/browse/B2CA-2132?focusedCommentId=553816)=> it appears to exists before the NBGL migration - let's consider it separately 
- [ ] Messages indexes restart on each chunk using `nbgl_useCaseReviewStreamingContinue()`  => we should fix it  probably by removing chunking and streaming, as separate change-set


src/boilerplate/io.c
```diff
@@ -46,6 +46,7 @@ struct {
 // set to true when the "Processing..." screen is shown, in order for the dispatcher to know if the
 // UX is not in idle state at the end of a command handler.
 bool G_was_processing_screen_shown;
+char const *G_processing_screen_text;
```
bigspider: Maybe worth a comment to explain that if set to NULL, a default message is shown (and otherwise it must be a pointer to a valid 0-terminated string).

src/handler/sign_message.c
```diff
@@ -33,6 +33,9 @@
 #define MAX_DISPLAYBLE_CHUNK_NUMBER \
     (5 * MESSAGE_CHUNK_PER_DISPLAY)  // If the message is too long we will not display it
 
+extern const char GA_LOADING_MESSAGE[];
+extern char const* G_processing_screen_text;
+
```
bigspider, line 37: I think it would be cleaner to have a `set_processing_screen_text(msg)` message in `display.c` (with header in `display.h`), so that all the places where it's used can just call the method and avoid seeing the `extern` variables.

Up to you if you prefer to declare the various `GA_` constants in a centralized place, or rather each module that wants to customize the message declares it in its own file.

bigspider (dismissed):
LGTM, only minor comments

src/boilerplate/io.c
iartemov-ledger: Fixed.

src/handler/sign_message.c
iartemov-ledger: Fixed.

bigspider (approved):

PR #335, opened 2025-06-16
Adaptation to ragger/pull/246 (blocking apdus revamp)

iartemov-ledger:
With this change we do not send 2nd identical APDU to Speculos anymore, but we rely on updated async Ragger API.

Depends on: https://github.com/LedgerHQ/ragger/pull/246


mruiz-ledger:
@iartemov-ledger we should add a commit on the PR updating the version of ragger

mruiz-ledger (approved):

PR #336, opened 2025-06-17
Removing unneeded sudo

iartemov-ledger:


tests/setup_script.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-sudo apt update -y && sudo apt install -y curl
+apt update -y && sudo apt install -y curl
```
Sjors, line 3: It seems strange that only the `sudo` for `apt update` isn't needed. At least on my desktop Ubuntu it _is_ needed generally.

If you're root than of course none of the sudo's are needed.

PR #346, opened 2025-07-30
Add pytest option to choose the speculos port

bigspider:
This allows running tests against a specific speculos instance that is already running, which is useful when debugging.


iartemov-ledger (approved):
Looks fine for me

PR #351, opened 2025-08-12
Allow custom app developer in UI for derived apps

bigspider:
Defaults to "Ledger" if not specified.


iartemov-ledger (approved):
Elegant.

PR #357, opened 2025-09-29
Harden unsafe function

bigspider:
The function `crypto_derive_symmetric_key` is dangerously written, as it has potential for buffer overflows. It is only used in a safe way with a fixed constant in the code, but it is better to be conservative and prevent possible future problems.

### DIFF
--- a/src/crypto.c
+++ b/src/crypto.c
@@ -275,11 +275,17 @@ uint32_t crypto_get_master_key_fingerprint() {
 }
 
 bool crypto_derive_symmetric_key(const char *label, size_t label_len, uint8_t key[static 32]) {
-    // TODO: is there a better way?
-    //       The label is a byte string in SLIP-0021, but os_derive_bip32_with_seed_no_throw
-    //       accesses the `path` argument as an array of uint32_t, causing a device freeze if memory
-    //       is not aligned.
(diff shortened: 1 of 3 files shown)


iartemov-ledger (approved):
Approving with 2 comments:
* the alignment requirement might be already obsolete after the sunset of Nano S:
    *  Nano S is Cortex-M0: https://developer.arm.com/documentation/dui0662/b/The-Cortex-M0--Instruction-Set/About-the-instruction-descriptions/Address-alignment
    * Nano X is Cortex-M3: https://developer.arm.com/documentation/dui0552/a/the-cortex-m3-instruction-set/about-the-instruction-descriptions/address-alignment  - unaligned access permitted for soem instructions if `UNALIGN_TRP` is not set.
    * Nano S+ and others are Cortex-M35P - seems to be the same as for Cortex-M3
 
* we might try to play with https://clang.llvm.org/docs/AttributeReference.html#align-value attribute

iartemov-ledger:
Already 79 "snapshot does not match" Apex Plus test failures :(
Not related  to your PR  - this should be an update in the SDK.

bigspider:
> * the alignment requirement might be already obsolete after the sunset of Nano S:

I tried on NanoS+, unfortunately it is still needed.

PR #358, opened 2025-10-01
Unstreaming for transaction send + Warning message unification

iartemov-ledger:
The goal - to cover more outputs (1 only output currently) with the classical review  rather when with the streaming.

## TODO/DONE

- [x] with the number of external outputs less or equal to **16**  we use classical `nbgl_useCaseReview()`
- [x]  all the warnings/security risks are shown using `nbgl_useCaseChoice()` for the both classical and streaming cases, so with possibility to interrupt if needed. "Non-default sighash", "external inputs" and "Unverified inputs" warnings are shown before "Review transaction to send Bitcoin"
- [x]  "High-fees" warning is shown as tag/value review "centeredInfo" item  (both classical and streaming cases)
- [x]  "Account name"  (streaming) or "From" (classical) information for transaction is shown as  tag/value review item
     - [x] finally "From" is set for the both classical and streaming for tx send.
- [x] For the status/nbgl_useCaseSpinner
     - [x]  for the transaction signature case it is shown as "Loading Transaction" (at the beginning and for each streaming loading) - > "Signing transaction"  -> "Transaction signed"
     - [x]  for the message signature case it is shown as "Loading message" (at the beginning and for each streaming loading)   - > "Message signed"   (not "Signing message" as it the signature is immediate) 

### Tests 
- [x]  16 and 17 external outputs
- [x]  maximum tests with 3 warnings and high-fees - all the warnings are covered. It was not possible by easy means to forge a transaction with all the 3 warnings. 
- [x]  all snapshots were removed and recreated from scratch using automatic snapshot generation and PR creation 
- [x] back direction for navigation to make sure that all the shown values remain
- [x] for 4to3, 4to17 and 4to18 sign_psbt_singlesig_wpkh test the address is different between my PC and GH (still the same between the runs), why :question:
    - it was because of a random (very bad one) used in `txmaker.py` - now it's fixed to a reproducible sequence of SHA256  calls. 

## Preliminary size impact estimation

For Flex target (DEBUG=0)
| Section | `develop` | this branch | 
| ----------- | --------------| -----------------|
| .text      | 0x17587  (95623₁₀)  |  0x17587 (95623₁₀)    |
| .bss       | 0x184c  (6220₁₀)     | 0x2694      (9876₁₀)|

**So ~3.6k of .bss increase - that is expected.  (we've got at minimum 28k on Nano X for the both .bss and stack, so the increase seems to be reasonable)**

### Stack consumption

For the following call stack in this branch:
```
48 bytes      #0  0x400125ee in nbgl_useCaseReview ()  
32 bytes      #1  0x4000bb86 in ui_display_transaction_simplified_flow_show ()
16 bytes      #2  0x4000b838 in ui_transaction_simplified_show ()
240 bytes     #3  0x40008374 in display_transaction ()
5112 bytes    #4  0x400077e0 in handler_sign_psbt ()
32 bytes      #5  0x40000118 in apdu_dispatcher ()
48 bytes      #6  0x4000a35a in app_main ()
80 bytes      #7  0x4000a476 in coin_main ()
8 bytes       #8  0x4000001e in main ()
```

**it gives the total of  5616 bytes of the stack consumption knowing that we've got 8k as the fixes OS stack size limitation**

in `develop` it is:
```
48 bytes     #0  0x40012442 in nbgl_useCaseReview ()
40 bytes     #1  0x4000b9e6 in ui_accept_transaction_simplified_flow ()
48 bytes     #2  0x4000b79a in ui_validate_transaction_simplified ()
248 bytes    #3  0x4000835e in display_transaction ()
3776 bytes   #4  0x400077ec in handler_sign_psbt ()
32 bytes     #5  0x40000118 in apdu_dispatcher ()
48 bytes     #6  0x4000a2fa in app_main ()
80 bytes     #7  0x4000a416 in coin_main ()
8 bytes      #8  0x4000001e in main ()
```
it gives the total of  4328 bytes of the stack consumption.

**So 1,288 bytes of increased stack consumption.**
Mainly this is because of increased size of `sign_psbt_state_t st` variable in `handler_sign_psbt()` function.


src/constants.h
```diff
@@ -36,6 +36,16 @@
...
 
+/**
+ * Maximum length of output index string ("101 of 203").
+ */
+#define MAX_OUTPUT_INDEX_LENGTH 12
+
+/**
+ * Maximum number of external outputs handled simultaneously.
+ */
+#define MAX_EXT_OUTPUT_NUMBER 10
```
bigspider: The app can do more, just not with the simplified flow. So maybe better a name like `MAX_EXT_OUTPUT_NUMBER_SIMPLIFIED`, and update the description accordingly.

bigspider: If it doesn't become a memory issue, increasing more (e.g. to 16 or even 32) might be worth it.
Not unthinkable that someone would do more than 10 payments in the same transaction (e.g.: a company paying salary to many employees).

Much larger numbers become unusable anyway, I guess.

src/handler/lib/policy.c
```diff
@@ -1542,6 +1542,11 @@ bool check_wallet_hmac(const uint8_t wallet_id[static 32], const uint8_t wallet_
     // to avoid timing-attack that could be exploited to extract it.
     result = os_secure_memcmp((void *) wallet_hmac, (void *) correct_hmac, 32) == 0;
 
+    PRINTF("%s: correct_hmac\n", __func__);
+    for (int i = 0; i < 32; i++) {
+        PRINTF("%02X", correct_hmac[i]);
+    }
+    PRINTF("\n");
```
bigspider: forgotten debug comments?

src/ui/display_nbgl.c
```diff
@@ -567,41 +574,32 @@ void ui_display_default_wallet_address_flow(void) {
...
+                       msg,
+#ifdef SCREEN_SIZE_WALLET
+                       GA_BACK_TO_SAFETY,
+                       GA_CONTINUE_ANYWAY,
+                       start_transaction_callback_inverted);
+#else
+                       GA_BACK_TO_SAFETY,
+                       GA_CONTINUE_ANYWAY,
+                       start_transaction_callback_inverted);
+#endif
```
bigspider: Aren't the `#if` and the `#else` branches identical?

src/constants.h
```diff
@@ -36,6 +36,16 @@
  */
 #define MAX_WALLET_NAME_LENGTH 64
 
+/**
+ * Maximum length of output index string ("101 of 203").
+ */
+#define MAX_OUTPUT_INDEX_LENGTH 12
```
bigspider: You could use `sizeof("101 of 203")` instead of the numeric constant.
Note that would be 11 and not 12 in this case (it includes the `\0`), but I suppose that's enough.

Might be useful to mention in the comment what other constant the string depends on (that is, why is it 203? Why it can't be 4 digits instead of 3? Can any assertion be added somewhere to guard from possible future off-by-one errors?).

src/handler/sign_psbt.h
```diff
@@ -102,10 +102,10 @@ typedef struct signing_state_s {
...
 } signing_state_t;
 
-// We cache the first 2 external outputs; that's needed for the swap checks
-// Moreover, this helps the code for the simplified UX for transactions that
-// have a single external output.
-#define N_CACHED_EXTERNAL_OUTPUTS 2
+// We cache the first MAX_EXT_OUTPUT_NUMBER external outputs;
+// This is used by the code for the simplified UX for transactions;
+// Moreover, that is needed for the swap checks.
+#define N_CACHED_EXTERNAL_OUTPUTS MAX_EXT_OUTPUT_NUMBER
```
bigspider: Being paranoid, it might be worth adding a `_Static_assert` in `execute_swap_checks` to make sure that `N_CACHED_EXTERNAL_OUTPUTS` is at least 2.

src/handler/sign_psbt.c
```diff
@@ -1383,58 +1383,70 @@ static bool __attribute__((noinline)) display_transaction(
...
-        if (!st->is_wallet_default && !ui_authorize_wallet_spend(dc, st->wallet_header.name)) {
+        // If it's not a default wallet policy, let's save this info to ask the user for
+        // confirmation
+        if (!st->is_wallet_default && !ui_authorize_wallet_spend(st->wallet_header.name)) {
             SEND_SW(dc, SW_DENY);
             return false;
         }
 
-        if (!display_warnings(dc, st)) {
+        // On NBGL devices, show the pre-approval screen
```
bigspider: By now, all devices are NBGL devices. So this should either be removed, or updated to reflect the correct conditional

src/ui/display.c
```diff
@@ -238,17 +238,17 @@ bool ui_display_wallet_address(dispatcher_context_t *context,
     return io_ui_process(context);
 }
 
-bool ui_authorize_wallet_spend(dispatcher_context_t *context, const char *wallet_name) {
-    ui_wallet_state_t *state = (ui_wallet_state_t *) &g_ui_state;
+bool ui_authorize_wallet_spend(const char *wallet_name) {
```
bigspider: Not a fan of having side effects (like storing something for later visualization) in the ui_ functions.

If it's easier to keep this kind of flow, let's find a different prefix for such functions. Perhaps a naming scheme like `prepare_ui_{what_the_function_does}`?

For this function, the return value could be simplified to `void`, updating the caller appropriately.

src/ui/display.c
```diff
@@ -238,17 +238,17 @@ bool ui_display_wallet_address(dispatcher_context_t *context,
...
+    ui_validate_transaction_state_t *state = (ui_validate_transaction_state_t *) &g_ui_state;
 
 #ifdef HAVE_AUTOAPPROVE_FOR_PERF_TESTS
     return true;
 #endif
```
bigspider: This is safe to delete now, since no UX is shown here.

src/ui/display_nbgl.c
```diff
@@ -116,6 +140,14 @@ static void start_transaction_callback(bool confirm) {
     }
 }
 
+static void start_transaction_callback_inverted(bool confirm) {
```
bigspider: A bit strange to have a function with this name. I suppose this comes from slightly 'abusing' `nbgl_useCaseChoice` in `ui_display_warning_generic`.

Perhaps cleaner to rename the callback to `display_warning_generic_callback` and move it right before  `ui_display_warning_generic`.

src/ui/display_nbgl.c
```diff
@@ -128,108 +160,69 @@ static void generic_content_callback(int token, uint8_t index, int page) {
...
-
 #define COMBINE(a, b) a b
 
 // create the string "0 <coind_id> (self-transfer)"
 #define SELF_TRANSFER_DESCRIPTION COMBINE("0 ", COMBINE(COIN_COINID_SHORT, " (self-transfer)"))
 
-void ui_accept_transaction_simplified_flow(void) {
-    _Static_assert(N_UX_PAIRS >= 9, "Insufficient pairs for this flow");
-
+void ui_accept_transaction_simplified_flow_init(void) {
```
bigspider: Same as `ui_authorize_wallet_spend`, this cold be a `prepare_ui_***` function.

src/ui/display_nbgl.c
```diff
@@ -128,108 +160,69 @@ static void generic_content_callback(int token, uint8_t index, int page) {
...
         };
     }
+}
 
-    if (!g_ui_state.validate_transaction_simplified.is_self_transfer) {
-        pairs[n_pairs++] = (nbgl_layoutTagValue_t){
-            .item = "Amount",
-            .value = g_ui_state.validate_transaction_simplified.amount,
-        };
+void ui_accept_transaction_simplified_flow_add(void) {
```
bigspider: Same as `ui_authorize_wallet_spend`, this cold be a `prepare_ui_***` function.

src/ui/display.h
```diff
@@ -194,15 +189,18 @@ bool ui_warn_high_fee(dispatcher_context_t *context);
...
-                                        uint64_t fee);
+void ui_validate_transaction_simplified_init(const char *wallet_policy_name,
+                                             unsigned int outputs_num,
+                                             tx_ux_warning_t warnings);
+void ui_validate_transaction_simplified_add(const char *coin_name,
+                                            uint64_t amount,
+                                            const char *address_or_description);
+bool ui_validate_transaction_simplified_start(const char *coin_name,
+                                              dispatcher_context_t *context,
+                                              uint64_t fee);
```
bigspider: While this whole header is generally not documented, these three functions need to be called in a specific order, so it might be worth documenting the correct usage (e.g.: first init, then add once or multiple times, then call _start).

src/ui/display.h
```diff
@@ -228,16 +226,17 @@ void ui_display_unverified_segwit_inputs_flows(void);
...
 void ui_warn_high_fee_flow(void);
 
 void ui_accept_transaction_flow(bool is_self_transfer);
 
 void ui_display_register_wallet_policy_flow(void);
-void ui_accept_transaction_simplified_flow(void);
+
+void ui_accept_transaction_simplified_flow_init(void);
+void ui_accept_transaction_simplified_flow_add(void);
+void ui_accept_transaction_simplified_flow_start(void);
```
bigspider: For future reference: I think the separation between these functions and their callers (`ui_validate_...`) is no longer useful, now that all implementations are using nbgl.

Lots of similar simplifications can probably be done while merging `display.c` and `display_nbgl.c` (and I guess `menu.c and `menu_nbgl.c`).

Probably better to keep it for a separate refactoring PR , though.

bigspider (changes requested):
Mostly nits here and there.

Suggested global improvements to the PR:
- <del>We should probably have at least one test (with ragger screenshots if it's not too burdensome) with `1 + MAX_EXT_OUTPUT_NUMBER` external outputs, so we cover the case where it's still using the streaming flow.</del> Nevermind, I found it - `test_sign_psbt_singlesig_wpkh_4to12`.
- use descriptive commit messages (can update with `rebase -i`), otherwise the git history becomes confusing after merging the PR.

src/handler/sign_psbt.h
iartemov-ledger: There is already this one https://github.com/LedgerHQ/app-bitcoin-new/blob/ui_alignment/src/handler/sign_psbt.c#L1144, it is not enough ?

bigspider: Indeed, I think that's good enough.

src/constants.h
iartemov-ledger: I've set 16, more it's really to much to go through manually. I will update the consumption info in the title.

src/handler/lib/policy.c
iartemov-ledger: Removed.

src/ui/display_nbgl.c
iartemov-ledger: Finally not. 
The Product indeed want to invert the warning buttons. I looks OK for touch screen devices:
image

but for Nano ones it would be weird with :heavy_check_mark: and :x: signs.

src/constants.h
iartemov-ledger: Done !

src/handler/sign_psbt.c
iartemov-ledger: Removed.

src/ui/display.c
iartemov-ledger: Removed where no UI is invoked.

src/ui/display_nbgl.c
iartemov-ledger: Explained here: https://github.com/LedgerHQ/app-bitcoin-new/pull/358#discussion_r2501966572
Done as you proposed.

src/ui/display.c
iartemov-ledger: Done for this particular function renaming it to `ui_prepare_authorize_wallet_spend()`.
See below for the init, add, show functions.

src/ui/display_nbgl.c
iartemov-ledger: I agree for the `ui_prepare_authorize_wallet_spend()` function.

But here I believe the init, add, show functions are self-documented.
I'm proposing the following naming (mixing historical naming and new realities) :
* In `display.c`:
    * For the simplified/classical case (`ui_transaction_simplified_` prefix):
        * `ui_transaction_simplified_init`
        * `ui_transaction_simplified_add`
        * `ui_transaction_simplified_show` (only this one shows something, the 2 above do some preparations)
    * for the streaming case (the prefix updated to ` ui_transaction_streaming_`):
        * `ui_transaction_streaming_prompt`
        * `ui_transaction_streaming_validate_output`
        * `ui_transaction_streaming_validate`
* In `display_nbgl.c`:
    * For the simplified/classical case (`ui_display_transaction_simplified_flow_` prefix):
        * `ui_display_transaction_simplified_flow_init`
        * `ui_display_transaction_simplified_flow_add`
        * `ui_display_transaction_simplified_flow_show`
    * for the streaming case (the prefix updated to ` ui_display_transaction_streaming_`):
        * `ui_display_transaction_streaming_prompt`
        * `ui_display_transaction_streaming_output_address_amount`
        * `ui_display_transaction_streaming_flow`

It is document here: https://github.com/LedgerHQ/app-bitcoin-new/blob/ui_alignment/src/ui/display.h#L182.

I  had understood the difference between display.c and `display_nbgl.c` c as follows:
* `display.c` - prepares the UI context
* `display_nbgl.c` - prepare the NBGL context  + actually displays the screen
But I agree that we can reconsider and maybe merge the both. Let's indeed postpone it to a separate PR.

src/ui/display_nbgl.c
iartemov-ledger: See https://github.com/LedgerHQ/app-bitcoin-new/pull/358#discussion_r2502052261.

src/ui/display.h
iartemov-ledger: Done. See https://github.com/LedgerHQ/app-bitcoin-new/pull/358#discussion_r2502052261.

src/ui/display.h
iartemov-ledger: I agree, see https://github.com/LedgerHQ/app-bitcoin-new/pull/358#discussion_r2502052261.

ragger_bitcoin/ragger_instructions.py
```diff
@@ -43,12 +44,12 @@ def review_start(self, output_count: int = 1, save_screenshot=True, has_warning=
...
 
         if has_warning:
-            self.same_request("Warning", NavInsID.USE_CASE_REVIEW_TAP, NavInsID.USE_CASE_CHOICE_CONFIRM,
+            self.same_request("Security risk detected", NavInsID.USE_CASE_REVIEW_TAP, NavInsID.USE_CASE_REVIEW_TAP,
                             save_screenshot=save_screenshot)
 
         for output_index in range(0, output_count):
-            # the initial 2 outputs are cached; that depends on the N_CACHED_EXTERNAL_OUTPUTS constant
-            if output_index < 2:
+            # the initial 10 outputs are cached; that depends on the N_CACHED_EXTERNAL_OUTPUTS constant
```
bigspider:
```suggestion
            # the initial N_CACHED_EXTERNAL_OUTPUTS outputs are cached, so it is the same request
```

bigspider (dismissed):
LGTM, only a small nit

bigspider (approved):

PR #359, opened 2025-10-06
Snapshots (automatic) update and other fixes

iartemov-ledger:
- [x] automatic snapshots update in the build/test workflow
- [x] corresponding Apex Plus snapshots update (as the SDK has changed) 
- [x] fixing the tests with the Python library and legacy Bitcoin app (by finally freezing the Speculos version just before the breaking change)


.github/workflows/ci-workflow.yml
```diff
@@ -174,7 +174,7 @@ jobs:
     runs-on: ubuntu-latest
 
     container:
-      image: ghcr.io/ledgerhq/speculos:latest
+      image: ghcr.io/ledgerhq/speculos:0.25.5
```
bigspider, line 178: Might be worth adding a comment for future reference on why this is fixing the version

bigspider (dismissed):
LGTM, only a nit

.github/workflows/ci-workflow.yml
iartemov-ledger: Good point ! Fixed !

bigspider (approved):

PR #372, opened 2025-11-13
Unstream message signing

iartemov-ledger:
- [x] code
   - [x]  only classical NBGL API is used for the message signatures
   - [x]  if fits (640 bytes currently) the message is copied by chunks to a local buffer and displayed, otherwise the message hash is displayed
   - [x]  only one single Merkle tree verification for a message (still per chunk, but there is no second pass) 
   - [x]  this is NBGL that handles the message split per screen 
         - [x] on wide-screen device if a message does not fit a screen it can be seen by clicking the "More button" 
         - [x] on Nano device we scroll through all the length (~10 screen for the 640 bytes message)
- [x] test adaptation
- [x] snapshots update


tests/instructions.py
```diff
@@ -24,13 +25,26 @@ def message_instruction_approve_long(model: Firmware) -> Instructions:
...
-        instructions.new_request("Loading message")
-        instructions.new_request("Loading message")
-        instructions.new_request("Loading message")
-        instructions.new_request("Sign message")
+        instructions.same_request("Sign message")
     else:
-        instructions.review_message(page_count=5)
+        # TODO: to wrap below to review_message_long() and
+        # to use coordinates from positions.py when available in Ragger
+        if (model.name in ["apex_p", "appex_m"]):
```
bigspider: typo, `appex_m` ==> `apex_m`.
Maybe we should just have `apex_p` anyway?

iartemov-ledger: Fixed !

bigspider (approved):
Please add an entry to the CHANGELOG so the improvements of this PR (and also the previously merged ones for multi-output transactions) will be in the release notes for the next version :)

PR #373, opened 2025-11-14
Switching to new Master Key Fingerprint Syscall +  Derivation Path Hardening

iartemov-ledger:
Depends on:
- https://github.com/LedgerHQ/ledger-secure-sdk/pull/1324 - for the wildcard path support
- https://github.com/LedgerHQ/speculos/pull/612 - for the testing
- https://github.com/LedgerHQ/ledger-app-database/pull/451.

Ad hoc testing using app-tester and hardened Speculos (https://github.com/LedgerHQ/speculos/tree/btc_hardening_not_mergeable) and a special app-tester branch (https://github.com/LedgerHQ/ledger-app-tester/tree/refs/heads/selected_apps_2): 
- https://github.com/LedgerHQ/ledger-app-tester/actions/runs/19663598985 :green_circle:


bigspider (changes requested):
Yup, it seems mostly that's what's needed.

The only recommended change would be to give an explicit error message if someone tries to get an xpub for a path that is now forbidden in `get_extended_pubkey`; while we don't expect people to be using them, paths like `m`, `m/something`, `m/something'` where previously allowed if `display=True`, and now need to be forbidden. in all cases

src/crypto.c
```diff
@@ -339,14 +349,39 @@ int get_extended_pubkey_at_path(const uint32_t bip32_path[],
...
+            case 0x4212: {
+                PRINTF(
+                    "Attempt to derive a key at root level without "
+                    "HAVE_APPLICATION_FLAG_DERIVE_MASTER permission.\n");
+                break;
+            }
+            case 0x4215: {
+                PRINTF("Attempt to derive a key at unauthorized path.\n");
+                break;
+            }
```
bigspider: Is there no constants we can import from the SDK for this? It seems a bit frail to have hardcoded OS constants.

src/crypto.c
```diff
@@ -339,14 +349,39 @@ int get_extended_pubkey_at_path(const uint32_t bip32_path[],
...
+            case 0x4214: {
+                PRINTF("Attempt to derive with incorrect curve.\n");
+                break;
+            }
```
bigspider: Since the Bitcoin app should never derive with the wrong curve, I think it's better to delete this case and let it go to the `default` case and make it return BAD_STATE instead of NOT_SUPPORTED.

src/crypto.c
```diff
@@ -339,14 +349,39 @@ int get_extended_pubkey_at_path(const uint32_t bip32_path[],
...
+            default: {
+                sw = SW_BAD_STATE;
+                PRINTF("Failed getting bip32 pubkey, error = 0x%08X\n", error);
+            }
+        }
+    }
+    return sw;
```
bigspider: The function is generally hard to read. I think it would be better to delete the `sw` variable and replace every assignment with early `return`s, which also reduces nesting. Something like:

```C
uint16_t cx_err_to_sw(cx_err_t error) {
    if (error == CX_OK) {
        return SW_OK;
    }
    switch (error) {
        case 0x4212: {
                PRINTF(
                    "Attempt to derive a key at root level without "
                    "HAVE_APPLICATION_FLAG_DERIVE_MASTER permission.\n");
                break;
            }
        // etc
        default: {
            PRINTF("Failed getting bip32 pubkey, error = 0x%08X\n", error);
            return SW_BAD_STATE;
        }
    }
    return SW_NOT_SUPPORTED;
}
```

bigspider (changes requested):
Only some comments on the `cx_err_to_sw` function, otherwise LGTM.

src/crypto.c
iartemov-ledger: For me it has to be propagated to the SDK by the OS team, as the values are returned to the user space and the apps should be able to handle them, may be at least as some abstract values, but today it is not the case.
Today it is defined in an internal OS header.

src/crypto.c
iartemov-ledger: Done!

src/crypto.c
iartemov-ledger: Simplified!

src/crypto.c
bigspider: Fine to leave a comment, we can update if/when the constants are exposed more cleanly.

bigspider (approved):

PR #374, opened 2025-11-14
Removing the special handling and the warning for the suspicious derivation paths

iartemov-ledger:
The new  `handler_get_extended_pubkey` behavior:
| `display` flag   | `path` detected as suspicious  | Result  |
|---|---|---|
| False  | False | Public key is returned without displaying  |
| False  | True  | `SW_NOT_SUPPORTED` returned  |
| True  | False  |  Normal  display public key flow  (based on `nbgl_useCaseReviewLight()`) and public key is returned |
| True  | True   |  **No warning**, normal  display public key flow  (based on `nbgl_useCaseReviewLight()`) and public key is returned - **the user will see the path and the key anyway, no need to warn in a special way**  |


bigspider (approved):
LGTM

PR #377, opened 2025-11-17
Fix speculos-bitcoin image in CI

### LINKED ISSUE #312: Use `speculos-bitcoin` image for e2e tests once possible
e2e tests are currently using `ghcr.io/ledgerhq/app-bitcoin-new/speculos-bitcoin-musig2`, which is created in [a custom branch](https://github.com/LedgerHQ/app-bitcoin-new/tree/speculos-bitcoin-with-musig2).

Once musig is merged on bitcoin (PR [here](https://github.com/bitcoin/bitcoin/pull/29675)), we should switch back to the regular `speculos-bitcoin` image.

bigspider:
Fix the build of the `speculos-bitcoin` container image (that was missing some dependencies), and stops using the `speculos-bitcoin-musig2` image in the CI: that is no longer necessary since MuSig2 support is merged in [bitcoin-core](https://github.com/bitcoin/bitcoin) master.

It also fixes the usage of the rpc in e2e tests, using the `getbalances` RPC in replacement of the `getwalletinfo` rpc (whose 'balance' field has been removed).

Closes: #312

### DIFF
--- a/.github/workflows/Dockerfile
+++ b/.github/workflows/Dockerfile
@@ -21,8 +21,11 @@ RUN cd / && \
 FROM ghcr.io/ledgerhq/speculos:latest
 COPY --from=0 /usr/local/bin/ /usr/local/bin/
 
+# install essential tools
+RUN apt update -y && apt install -y build-essential curl git
+
 # install runtime dependencies for bitcoind
-RUN apt update -y && apt install -y libminiupnpc-dev libminiupnpc-dev libnatpmp-dev libevent-dev libzmq3-dev
+RUN apt install -y libminiupnpc-dev libminiupnpc-dev libnatpmp-dev libevent-dev libzmq3-dev
 
 # Add bitcoin binaries to path
 ENV PATH=/usr/local/bin/:$PATH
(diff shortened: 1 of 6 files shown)


.github/workflows/Dockerfile
```diff
@@ -13,16 +13,19 @@ RUN apt install -y bsdmainutils build-essential cmake pkg-config ccache git libb
...
+    cmake -B build -DENABLE_IPC=OFF && \
     cmake --build build && \
     cmake --install build
 
 
 FROM ghcr.io/ledgerhq/speculos:latest
 COPY --from=0 /usr/local/bin/ /usr/local/bin/
 
+# install essential tools
+RUN apt update -y && apt install -y build-essential curl git
```
iartemov-ledger, line 25: Is this needed as well to run `bitcoind`  ?

iartemov-ledger (approved):
I dreamt of  this :) 

The container sha256:9c01833661651985363ce8a28da97a9e0c932f37b910b9cf55e78fdf2d55ec58 used for tests corresponds to the latest one deployed https://github.com/LedgerHQ/app-bitcoin-new/pkgs/container/app-bitcoin-new%2Fspeculos-bitcoin/579404333?tag=latest.

Just one quick question in the PR.

.github/workflows/Dockerfile
bigspider: This is from multiple iterations of the container failing to pass the CI _after_ building it for 40m...
- `git` seems to no longer be in the base `speculos` image (I guess the base container was changed in the last few months?), and without that the github action runner just fails completely.
- `build-essential` is needed because otherwise one of the python dependencies used in the python tests fails to build
- `curl` is probably not needed, but it's a tiny binary so I didn't care about testing it :P

PR #381, opened 2025-11-19
A couple of fixes in crypto.c

iartemov-ledger:
- [x] overflowed (in case of syscall exception) array with too small size 
- [x] uninitialized variable 
- [x] Stax/Flex snapshot linked strictly to the migration to API_LEVEL_25


src/crypto.c
```diff
@@ -289,16 +289,21 @@ bool crypto_derive_symmetric_key(const char *label, size_t label_len, uint8_t ke
...
-                                           key,
+                                           tmp_key,
                                            NULL,
                                            NULL,
                                            0) != CX_OK) {
         return false;
     }
+    /* It is assumed that with chain_code and seed equal to NULL the output key length is 32 bytes
+     */
+    memcpy(key, tmp_key, 32);
```
bigspider: `tmp_key` contains private key material, so it would be better to zero it with `explicit_bzero` before returning.
A slight refactoring might be needed to make sure that it's also zeroed in case `os_derive_bip32_with_seed_no_throw` fails (otherwise `return false` would skip the `explicit_bzero` call).

bigspider (changes requested):
Just a slight defensive hardening in the handling of the key.

src/crypto.c
iartemov-ledger: Good point, fixed !

src/crypto.c
```diff
@@ -289,18 +292,25 @@ bool crypto_derive_symmetric_key(const char *label, size_t label_len, uint8_t ke
...
+                                                      (uint32_t *) label_copy,
+                                                      label_len,
+                                                      tmp_key,
+                                                      NULL,
+                                                      NULL,
+                                                      0);
+    if (ret == CX_OK) {
+        /* It is assumed that with chain_code and seed equal to NULL the output key length is 32
+         * bytes
+         */
```
bigspider: nit: I should have commented this before, but I think for SLIP21 the last 32 bytes are just unused. I'd assume for BIP32 there are some curves where private keys are 64 bytes - in fact, for the secp256k1 curve used in bitcoin, private keys are also 32 bytes, so a 32 byte parameter doesn't make much sense.

We can just say in the comment that 'only the first 32 bytes are used for SLIP21', leaving the reason generic.

bigspider (dismissed):
LGTM, only a nit in the comment

src/crypto.c
iartemov-ledger: Addressed!

bigspider (approved):

PR #391, opened 2025-12-05
To fit key info

iartemov-ledger:
- [x] rearranging the key information so it fits without ...:
    - `Dummy key@1 (2/3)`
    - `Our key@0 (1/3)`
    - `Their key@2 (3/3)`
- [x] corresponding snapshots update
- [x]  version bump not needed as it will make part of v2.4.3


src/ui/display.c
```diff
@@ -150,35 +150,25 @@ bool ui_display_register_wallet_policy(
...
-                         sizeof(state->keys_label[i]),
-                         "Key @%u, %s",
-                         i,
-                         labels[2]);
+                num = 2;
                 break;
             default:
                 LEDGER_ASSERT(false, "Unreachable code");
         }
+        snprintf(state->keys_label[i], sizeof(state->keys_label[i]), "%s key %u", labels[num], i);
```
bigspider:
```suggestion
        snprintf(state->keys_label[i], sizeof(state->keys_label[i]), "%s key @%u", labels[num], i);
```

I would keep the `@`, because `@0`, `@1`, etc. refers to the corresponding numbers seen in the descriptor template just before. So it might be confusing to write `@2` in the descriptor template, and just `2` here. Hopefully it still fits?

bigspider (changes requested):
Only a small nit (assuming that the extra character is not problematic).

src/ui/display.c
iartemov-ledger: Finally that did not work well - the "Dummy" text was still too long:
Dummy_with_@

Finally I've come up with following version (no white-space between key and @):
`Dummy key@1 (1/3)`

Dummy_with_@_without_whitespace

What do you think ?

bigspider: I don't love it, but better than truncating, so let's fix it this way for this release.

This will still truncate with more than 9 keys - which is probably rare in practice, but we support it.
So perhaps let's track in a new issue that there's still room for improvement (the ideal would probably be to remove the `(i/n)` part entirely, if the SDK can make it happen).

bigspider (approved):
Left a comment - good fix, but let's further improve in the future if possible!

src/ui/display.c
iartemov-ledger: The issue created: https://github.com/LedgerHQ/app-bitcoin-new/issues/394.

PR #418, opened 2026-01-21
Migration to SDK/lib_standard_app [step 1]

iartemov-ledger:
- [x]  duplicated with SDK/lib_standard_app` modules removed
- [x]  the ones with added values, but colliding name renamed to `_ext`
- [x] Splitting up MAX_BIP388_XPUB_DERIVATION_STEPS (=8) and MAX_BIP32_PATH_STEPS (+2)
- [x]  header files inclusion is reorganized with Claude Haiku 4.5 as follows:
```
- C system headers using <>
- Empty line
- Unit header (for .c files only)
- Empty line + /* SDK headers */ section using ""
- Empty line + /* Local headers */ section using ""
- All headers alphabetically sorted within each category
- No folder paths - only header file names
- No leading empty lines
```
- [x] some unit tests removed as duplicated with the SDK, others updated


src/common/bip32_ext.h
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Index of first hardened child according to BIP32; it can also be used as the bitmask for hardened
+ * children.
+ */
+#define BIP32_FIRST_HARDENED_CHILD 0x80000000
+
+
+#define MAX_BIP44_ACCOUNT_RECOMMENDED       100
+#define MAX_BIP44_ADDRESS_INDEX_RECOMMENDED 50000
```
bigspider: Since we have a `constants.h` file, that seems like a good place for these, so this file can be deleted altogether.

Makefile
```diff
@@ -175,9 +175,12 @@ endif
```
bigspider, line 175: Typo `usigng` in commit message

src/common/bip32_ext.h
```diff
@@ -1,10 +1,12 @@
 #pragma once
 
+#include "lib_standard_app/bip32.h"
+
 /**
  * Maximum length of BIP32 path supported.
  * Note: BIP32 allows up to 256 derivation steps - but only 5 or 6 are used in most cases.
  */
-#define MAX_BIP32_PATH_STEPS 8
+#define MAX_BIP32_PATH_STEPS MAX_BIP32_PATH
```
bigspider: Not sure about this change.
- I think `lib_standard_app/bip32` shouldn't have an opinion about the number of derivation steps - that's a decision of apps. IMHO `MAX_BIP32_PATH` should actually be deleted from `lib_standard_app`.
- There is a reason there are (or were) two constants `MAX_BIP32_PATH_STEPS` and `MAX_BIP32_PATH`: one indicates how many derivation steps are OK for a path used in a wallet policy. But the actual keys (for example as they appear in the PSBT) will have 2 more derivation steps (because of the additional `/change/addr_index` part. So there is the need of distinguishing these two quantities.

The name of the constants was indeed not great, and the bigger one should be computer as "smaller one + 2", and moved in the same place.

src/handler/sign_psbt/txhashes.c
```diff
@@ -28,7 +28,7 @@ static void cb_process_data(buffer_t *data, void *cb_state) {
     callback_state_t *state = (callback_state_t *) cb_state;
 
     size_t data_len = data->size - data->offset;
-    uint8_t *data_start_ptr = data->ptr + data->offset;
+    uint8_t *data_start_ptr = (uint8_t *) data->ptr + data->offset;
```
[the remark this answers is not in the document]
iartemov-ledger: Yes, this is the declaration from the SDK, what can I do?

src/common/buffer_ext.c
```diff
@@ -0,0 +1,140 @@
...
+    return true;
+}
+
+bool buffer_write_u16(buffer_t *buffer, uint16_t value, endianness_t endianness) {
+    if (!buffer_can_read(buffer, 2)) {
+        return false;
+    }
+
+    if (endianness == BE) {
+        write_u16_be((uint8_t *) buffer->ptr, buffer->offset, value);
```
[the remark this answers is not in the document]
iartemov-ledger: Yes, this is the declaration from the SDK, what can I do?

src/common/buffer_ext.c
```diff
@@ -0,0 +1,140 @@
...
+    return true;
+}
+
+bool buffer_write_u64(buffer_t *buffer, uint64_t value, endianness_t endianness) {
+    if (!buffer_can_read(buffer, 8)) {
+        return false;
+    }
+
+    if (endianness == BE) {
+        write_u64_be((uint8_t *) buffer->ptr, buffer->offset, value);
```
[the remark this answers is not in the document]
iartemov-ledger: Yes, this is the declaration from the SDK, what can I do?

src/common/buffer_ext.c
```diff
@@ -0,0 +1,140 @@
...
+
+    return true;
+}
+
+bool buffer_write_bytes(buffer_t *buffer, const uint8_t *data, size_t n) {
+    if (!buffer_can_read(buffer, n)) {
+        return false;
+    }
+
+    memmove((uint8_t *) (buffer->ptr + buffer->offset), data, n);
```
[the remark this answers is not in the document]
iartemov-ledger: Yes, this is the declaration from the SDK, what can I do?

src/swap/handle_swap_sign_transaction.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+/* Local headers */
 #include "swap_lib_calls.h"
```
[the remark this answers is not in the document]
iartemov-ledger: It is provided by the SDK.

Makefile
iartemov-ledger: Fixed.

src/common/bip32_ext.h
iartemov-ledger: Fixed.

src/common/bip32_ext.h
iartemov-ledger: Addressed.

tests/test_sign_psbt.py
```diff
@@ -747,6 +749,115 @@ def test_sign_psbt_fail_wrong_non_witness_utxo(navigator: Navigator, firmware: F
...
+
+    # This should succeed as we're at the limit
+    result = client.sign_psbt(psbt, wallet, wallet_hmac, navigator,
+                              instructions=sign_psbt_instruction_approve(firmware, save_screenshot=False),
+                              testname=test_name)
+
+    assert len(result) == 1
+
+
+def test_sign_psbt_fail_too_deep_derivation_path(navigator: Navigator, firmware: Firmware, client:
```
bigspider: I'm not sure this test is useful - since registration of such policies is already forbidden at registration time. I'd delete this (but great to have the other tests!).

bigspider (commented):
Great work!

Maybe we could do all these new tests without saving snapshots, though. It doesn't seem to add much value and we already have too many!

tests/test_sign_psbt.py
iartemov-ledger: Removed.

bigspider (approved):

PR #425, opened 2026-02-18
Bug fixes 2026-02-18

iartemov-ledger:
- [x] Fixing call_get_merkleized_map_value() return value checking
- [x] Zeroing out nonce-related stack buffers in musig part
- [x] Fix partial_data_len=0 case

AI-generated unit tests with a small added value, but with a lot of code lines have been removed finally.

### DIFF
--- a/src/handler/sign_psbt.c
+++ b/src/handler/sign_psbt.c
@@ -278,7 +278,7 @@ init_global_state(dispatcher_context_t *dc, sign_psbt_state_t *st) {
                                                    1,
                                                    raw_result,
                                                    sizeof(raw_result));
-        if (result_len == -1) {
+        if (result_len < 0) {
             st->locktime = 0;
         } else if (result_len != 4) {
             SEND_SW(dc, SW_INCORRECT_DATA);
@@ -962,7 +962,7 @@ preprocess_outputs(dispatcher_context_t *dc,
                                                    output.in_out.scriptPubKey,
                                                    sizeof(output.in_out.scriptPubKey));
 
-        if (result_len == -1 || result_len > (int) sizeof(output.in_out.scriptPubKey)) {
+        if (result_len < 0 || result_len > (int) sizeof(output.in_out.scriptPubKey)) {
             SEND_SW(dc, SW_INCORRECT_DATA);
             return false;
         }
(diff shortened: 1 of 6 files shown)


src/handler/sign_psbt.c
```diff
@@ -278,7 +278,7 @@ init_global_state(dispatcher_context_t *dc, sign_psbt_state_t *st) {
                                                    1,
                                                    raw_result,
                                                    sizeof(raw_result));
-        if (result_len == -1) {
+        if (result_len < 0) {
             st->locktime = 0;
```
iartemov-ledger, line 282: In the case of negative return value would it be more correct to do `SEND_SW(dc, SW_INCORRECT_DATA);` 
as we do for all other occurrences.
Or here we do insist on `st->locktime = 0;` ?
@bigspider ?

bigspider: No, unfortunately there's no way to distinguish "the field is missing" from "an error happened". For fields that are not defined in the psbt (because they're not mandatory per the [BIP-174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki) specifications), treating absence as an error would reject valid transactions.

src/musig/musig.c
```diff
@@ -264,27 +264,32 @@ int musig_nonce_gen(const uint8_t *rand,
...
+    if (0 > crypto_get_compressed_pubkey(R_s1.raw, pubnonce->R_s1)) goto nonce_gen_fail;
+    if (0 > crypto_get_compressed_pubkey(R_s2.raw, pubnonce->R_s2)) goto nonce_gen_fail;
 
     return 0;
+
+nonce_gen_fail:
+    explicit_bzero(secnonce->k_1, sizeof(secnonce->k_1));
+    explicit_bzero(secnonce->k_2, sizeof(secnonce->k_2));
+    return -1;
 }
```
bigspider, line 293: This isn't really a vulnerability because in case of musig2 signing failures, the signing is invalidated and it has to start over from Round 1. So we'd be leaving in memory a 'secret' that is just a randomly generated number that is no longer useful for anything - a new attempt at signing will generate a new random number.
However, I guess the hardening doesn't hurt.

In that case, however, we should therefore also do the same in the callers of `musig_nonce_gen`:
- `produce_and_yield_pubnonce` doesn't need the secnonce at all, so it can delete it straight away
- `sign_sighash_musig_and_yield` should delete it once done (both on success and error cases).

iartemov-ledger: * Added immediate `secnonce` zeroing out in `produce_and_yield_pubnonce()`
* In `sign_sighash_musig_and_yield()` if I'm reading correctly that was already the case:
    * `secnonce` iz zeroed out in case of `musig_nonce_gen()` error 
    * it is zeroed out right after the latest use as well

bboilot-ledger (approved):

bigspider (approved):

PR #430, opened 2026-02-25
refactor(bitcoin_client_js): migrate descriptors to v3 and upgrade bitcoinjs-lib to v7

landabaso:
This PR is a follow up to #428. It enables descriptors v3 support and aligns dependencies with the current `bitcoinjs-lib` major.

- Migrate third-party address validation to `@bitcoinerlab/descriptors` v3 (`DescriptorsFactory` + `Output`), enabling tapscript address validation.
- Extend `appClient` wallet-address integration vectors with additional testnet descriptor/address cases from Python coverage.
- Upgrade `bitcoinjs-lib` from `6.x` to `7.x`
- Upgrade the TS/JS toolchain to current versions (`TypeScript 5.9`, `Jest 30`, `ts-jest`, `babel-jest`, `typedoc 0.28`) and set `node >= 18.14` (required by bitcoinjs-lib / @bitcoinerlab/descriptors).
- Keep the library API non-breaking by converting v7 `Uint8Array`/`bigint` values internally to existing `Buffer`/`number` interfaces.
- A future major release can switch public return types to Uint8Array/bigint if desired.

### DIFF
--- a/bitcoin_client_js/src/__tests__/appClient.test.ts
+++ b/bitcoin_client_js/src/__tests__/appClient.test.ts
@@ -104,7 +104,7 @@ describe("test AppClient", () => {
       repoRootPath + "/bin/app.elf",
       '--display', 'headless'
     ]);
-    
+
     sp.stdout.on('data', function(data) {
       if (process.env.LOG_SPECULOS) {
         console.log('stdout: ' + data);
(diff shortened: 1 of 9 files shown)


bitcoin_client_js/src/__tests__/appClient.test.ts
```diff
@@ -230,11 +230,101 @@ describe("test AppClient", () => {
         expResult: "tb1qmyauyzn08cduzdqweexgna2spwd0rndj55fsrkefry2cpuyt4cpsn2pg28",
         walletHmac: Buffer.from("d7c7a60b4ab4a14c1bf8901ba627d72140b2fb907f2b4e35d2e693bce9fbb371", "hex")
       },
+      // additional descriptor/address vectors from python tests (not previously covered in JS)
```
bigspider: Thanks for adding these tests!
I think this comment should be removed, as it is relevant in the PR, but not in the source code per se.

bigspider (dismissed):
Thanks for modernizing the JS client!

Only a small nit, otherwise LGTM.

bigspider (approved):

PR #433, opened 2026-03-02
Reject dangerous relative timelocks

bigspider:
Prevents the user from registering wallet policies that contains an `older(n)` fragment where `n` has non-zero bits in position that have no consensus meaning per [BIP-68](https://github.com/adewaleijalana/bitcoin-bips/blob/master/bip-0068.mediawiki)/[BIP-112](https://github.com/adewaleijalana/bitcoin-bips/blob/master/bip-0112.mediawiki).

For example, this avoids mistakenly registering something like `older(65536)` which effectively has no timelock.

### DIFF
--- a/src/common/wallet.c
+++ b/src/common/wallet.c
@@ -2906,6 +2906,125 @@ int compute_miniscript_policy_ext_info(const policy_node_t *policy_node,
     }
 }
 
+static int traverse_policy_node_tree(const policy_node_tree_t *tree,
+                                     policy_node_callback_t callback,
+                                     void *callback_state) {
+    if (tree->is_leaf) {
+        return traverse_policy_dfs(r_policy_node(&tree->script), callback, callback_state);
+    } else {
+        int ret = traverse_policy_node_tree(r_policy_node_tree(&tree->left_tree),
+                                            callback,
+                                            callback_state);
+        if (ret < 0) return ret;
+        return traverse_policy_node_tree(r_policy_node_tree(&tree->right_tree),
+                                         callback,
+                                         callback_state);
+    }
+}
+
+int traverse_policy_dfs(const policy_node_t *policy_node,
+                        policy_node_callback_t callback,
+                        void *callback_state) {
+    if (callback == NULL) {
+        return -1;
+    }
+
+    // Visit the current node first (pre-order)
+    int ret = callback(policy_node, callback_state);
+    if (ret < 0) return ret;
+
+    switch (policy_node->type) {
+        // Leaf nodes with no child scripts
+        case TOKEN_0:
+        case TOKEN_1:
+        case TOKEN_PK_K:
+        case TOKEN_PK_H:
+        case TOKEN_PK:
+        case TOKEN_PKH:
+        case TOKEN_WPKH:
+        case TOKEN_OLDER:
+        case TOKEN_AFTER:
+        case TOKEN_SHA256:
+        case TOKEN_HASH256:
+        case TOKEN_RIPEMD160:
+        case TOKEN_HASH160:
+        case TOKEN_MULTI:
+        case TOKEN_MULTI_A:
+        case TOKEN_SORTEDMULTI:
+        case TOKEN_SORTEDMULTI_A:
+            return 0;
+
+        // Nodes with a single child script (including miniscript wrappers)
+        case TOKEN_SH:
+        case TOKEN_WSH:
+        case TOKEN_A:
+        case TOKEN_S:
+        case TOKEN_C:
+        case TOKEN_T:
+        case TOKEN_D:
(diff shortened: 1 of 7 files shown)


src/handler/lib/policy.c
```diff
@@ -2069,6 +2088,20 @@ int is_policy_sane(dispatcher_context_t *dispatcher_context,
             }
         }
     }
+
+    // For relative timelocks with `older(n)`, bits 16 to 21 and 23 to 31 have no consensus meaning
+    // per BIP-68/BIP-112. Therefore, for example, `older(65536)` is equivalent to `older(0)`, which
```
iartemov-ledger, line 2090: There is a redundancy for the description here and above - better to keep only one explanation?

src/handler/lib/policy.c
```diff
@@ -2069,6 +2088,20 @@ int is_policy_sane(dispatcher_context_t *dispatcher_context,
...
+    // could be dangerous as the user might expect that a timelock is enforced. For relative
+    // timelocks, we reject any value outside the following safe ranges:
+    // - between 1 and 65535 (inclusive), if bit 22 is cleared (block-based timelock)
```
iartemov-ledger, line 2093: Also the description seems to be a bit ambiguous. 
Don't I understand correctly that only 0 - 15 bits are valid in the both  Block-based and Time-based cases. If yes, maybe just to say this, because  "1 + 2^22 = 4194305 and 65535" looks bit complex? 

Also you do not speak about `SEQUENCE_LOCKTIME_DISABLE_FLAG`.
OK, with `if (n < 1 || n > 65535)` we refuse it correctly, but is it yet worth mentioning?

src/common/wallet.c
```diff
@@ -2906,6 +2906,125 @@ int compute_miniscript_policy_ext_info(const policy_node_t *policy_node,
...
+            policy_node_scriptlist_t *cur = r_policy_node_scriptlist(&node->scriptlist);
+            while (cur != NULL) {
+                ret = traverse_policy_dfs(r_policy_node(&cur->script), callback, callback_state);
+                if (ret < 0) return ret;
+                cur = r_policy_node_scriptlist(&cur->next);
+            }
+            return 0;
+        }
+
+        case TOKEN_TR: {
```
iartemov-ledger, line 3013: Missed comment for the tree node type ? ;)

src/handler/lib/policy.c
```diff
@@ -2069,6 +2088,20 @@ int is_policy_sane(dispatcher_context_t *dispatcher_context,
...
+    // - between 1 + 2^22 = 4194305 and 65535 + 2^22 = 4259839 (inclusive), if bit 22 is set
+    // (time-based timelock)
+    //
+    // See also: https://github.com/bitcoin/bitcoin/pull/33135
+    if (0 > traverse_policy_dfs(policy, check_older_node_cb, NULL)) {
```
iartemov-ledger, line 2103: Why cannot we do the check in `read_and_parse_wallet_policy() -> parse_descriptor_template() -> parse_script()` where we already parse all token types, including `TOKEN_OLDER()` ?
Moreover `read_and_parse_wallet_policy()` is called even earlier than `is_policy_sane()` in `handler_register_wallet()`.

bigspider: It is technically valid miniscript that we're forbidding, so I think the `sanity` rules are a better place for this check, as they are there precisely to avoid footguns on otherwise valid scripts.

src/handler/lib/policy.c
bigspider: In 9db40d00ac1eaed7891c43bbf0e7cb2caf7cbebd, I consolidated all the comments in `is_policy_sane()`.

src/handler/lib/policy.c
bigspider: Basically it's bits 0 - 15, plus bit 22 that is the `SEQUENCE_LOCKTIME_TYPE_FLAG` - which is why the check in the callback first masks out bit 22, then simply checks the range [1, 65535].
[4194305, 4259839] is the range for time-based timelocks, and the comment tries to explain how the constants are computed (even if not directly used in the code - but it's used in the python tests).

I tried to make it more explicitly in 9db40d00ac1eaed7891c43bbf0e7cb2caf7cbebd.
I also added a comment about added a comment about why `SEQUENCE_LOCKTIME_DISABLE_FLAG` is irrelevant here - good to mention, indeed.

src/common/wallet.c
bigspider: Added in 9db40d00ac1eaed7891c43bbf0e7cb2caf7cbebd

iartemov-ledger (approved):

PR #446, opened 2026-03-23
Various small bug fixes and code improvements

bigspider:
Also a few code documentation improvements.


iartemov-ledger (approved):
A bit weird `size_t`, `int`, uint32_t` and `uint64_t` conversions in in `call_get_merkle_leaf_index()`, but finally  seems to be OK ;)

PR #448, opened 2026-03-24
Sighash gating

iartemov-ledger:
Described in https://ledgerhq.atlassian.net/browse/B2CA-2584.

**Implementation**
- [x] implementation (additional UI confirmation if user really allows non-standard signing rules)
- [x] a status window that is shown on non-standard sighash rejection informing about the new menu setting . The status disappears 3 seconds after (non-customizable) or if a user clicks on screen (wide-screen device) or presses 2 buttons (Nano devices).   

**Tests**
- [x] unit tests
- [x]  new functional settings test + update for sighash tests  
- [x] fixes for tests_mainnet
- [x] snapshot updates


unit-tests/test_sighash.c
```diff
@@ -0,0 +1,199 @@
...
+    assert_false(app_settings_get_allow_nondefault_sighash());
+
+    app_settings_set_allow_nondefault_sighash(true);
+    assert_true(app_settings_get_allow_nondefault_sighash());
+
+    app_settings_set_allow_nondefault_sighash(false);
+    assert_false(app_settings_get_allow_nondefault_sighash());
+}
+
+int main() {
```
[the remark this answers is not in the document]
iartemov-ledger: In all units test it is like this. Won't change.

src/ui/menu_nbgl.c
```diff
@@ -32,9 +38,84 @@ static const nbgl_contentInfoList_t infoList = {
...
+    if (token == ALLOW_NONDEFAULT_SIGHASH_TOKEN) {
+        if (!app_settings_get_allow_nondefault_sighash()) {
+            // About to enable: show a warning confirmation dialog
+            nbgl_useCaseChoice(&ICON_APP_WARNING,
+#ifdef SCREEN_SIZE_WALLET
+                               "Non-standard sighash",
+#else
+                               "Sighash types",
+#endif
+                               "Are you sure to allow\nnon-standard signing rules?",
```
bigspider: Maybe worth a more explicit version:

```suggestion
                               "This allows signing transactions that leave parts "
                               "unprotected. You'll still confirm each one.",
```

unit-tests/test_sighash.c
```diff
@@ -0,0 +1,172 @@
...
+
+static void test_sighash_default_is_safe_taproot(void **state) {
+    (void) state;
+    // SIGHASH_DEFAULT (0x00) is safe only for segwit version > 0 (Taproot)
+    assert_int_equal(classify_sighash(SIGHASH_DEFAULT, 1), SIGHASH_CLASS_SAFE);
+}
+
+static void test_sighash_default_is_not_safe_segwit_v0(void **state) {
+    (void) state;
+    // SIGHASH_DEFAULT for segwit v0 is unsupported, not safe
```
bigspider: `, not safe` can probably be removed, as it might be confusing with `SIGHASH_CLASS_NON_SAFE`.

unit-tests/test_sighash.c
```diff
@@ -0,0 +1,172 @@
...
+        cmocka_unit_test(test_sighash_anyonecanpay_all_is_unsupported_legacy),
+
+        // classify_sighash: UNSUPPORTED
+        cmocka_unit_test(test_sighash_0x80_alone_is_unsupported),
+        cmocka_unit_test(test_sighash_0x84_is_unsupported),
+        cmocka_unit_test(test_sighash_arbitrary_values_unsupported),
+    };
+
+    return cmocka_run_group_tests(tests, NULL, NULL);
+}
```
bigspider, line 104: It might be less verbose to have a single test that loops through a vector with all the testcases, where each testcase is a triple (sighash_type, segwit_version, classification_result). Also maybe easier to read.

(Feel free to ignore if you disagree.)

tests/test_sighash_setting.py
```diff
@@ -0,0 +1,301 @@
...
+from ragger.firmware import Firmware
+from ragger_bitcoin import RaggerClient
+
+from .conftest import toggle_nonstandard_sighash_setting
+from .instructions import sign_psbt_instruction_approve
+
+tests_root: Path = Path(__file__).parent
+
+# Error codes from error_codes.h
+EC_SIGN_PSBT_NONDEFAULT_SIGHASH_NOT_ALLOWED = 0x000d
```
bigspider, line 31: Might be for another PR, but it could be useful to have a generic helper to let python tests parse the constants directly from the `.h` headers instead of redefining.

bigspider (dismissed):
Looks good!
Some small suggestions.

It would be nice to have at least one screenshot that includes the rejection screen, but I guess ragger doesn't allow capturing it?

Alternatively, adding a few presets to the playground could be nice.

tests/test_sighash_setting.py
iartemov-ledger: It is possible, we need to bee carrefull because the `EC` constants are namespaced.
Better indeed to do it in a separate PR, created follow-up issue: https://github.com/LedgerHQ/app-bitcoin-new/issues/507.

unit-tests/test_sighash.c
iartemov-ledger: Good point, done!

unit-tests/test_sighash.c
iartemov-ledger: Fixed.

iartemov-ledger:
> It would be nice to have at least one screenshot that includes the rejection screen, but I guess ragger doesn't allow capturing it?

There are corresponding screenshots per device, see par example: https://github.com/LedgerHQ/app-bitcoin-new/blob/5578a782b69a75d735d98b14a7113aaf81456a4e/tests/snapshots/apex_p/test_sighash_none_rejected_by_default/00000.png.

src/ui/menu_nbgl.c
iartemov-ledger: Fixed

bigspider:
> There are corresponding screenshots per device, see par example: https://github.com/LedgerHQ/app-bitcoin-new/blob/5578a782b69a75d735d98b14a7113aaf81456a4e/tests/snapshots/apex_p/test_sighash_none_rejected_by_default/00000.png.

Ah, I was looking for this one but failed to find it, thanks!

bigspider (approved):
LGTM

PR #460, opened 2026-03-31
Tighten a few checks around type conversions.

bigspider:
Hardening and limit checks for a few type conversions

### DIFF
--- a/src/handler/sign_psbt/extract_bip32_derivation.c
+++ b/src/handler/sign_psbt/extract_bip32_derivation.c
@@ -1,5 +1,6 @@
 #include <stdint.h>
 #include <string.h>
+#include <limits.h>
 
 #include "extract_bip32_derivation.h"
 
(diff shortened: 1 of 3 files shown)


src/handler/sign_psbt/extract_bip32_derivation.c
```diff
@@ -21,7 +22,15 @@ typedef struct {
 } fpt_der_callback_data_t;
 
 static void fpt_der_data_len_callback(size_t data_length, void *callback_state) {
-    ((fpt_der_callback_data_t *) callback_state)->total_data_length = data_length;
+    fpt_der_callback_data_t *cs = (fpt_der_callback_data_t *) callback_state;
+
+    if (data_length > INT_MAX) {
+        // fail early if the conversion to int would overflow
+        // by setting the error result in the callback state
+        cs->result = -1;
```
bboilot-ledger, line 30: In this error case, do we also need to set `cs->total_data_length` to a specific value or is `cs->result` always checked first before any further processing?

bboilot-ledger (approved):

src/handler/sign_psbt/extract_bip32_derivation.c
bigspider: the second callback (the one receiving the actual data) always returns immediately if the `result` is set to a negative value:

https://github.com/LedgerHQ/app-bitcoin-new/blob/631cdf220e62c13b2e108861533cbc3b7864e9f2/src/handler/sign_psbt/extract_bip32_derivation.c#L36-L40

therefore, `total_data_length` is never accessed in that case.

PR #465, opened 2026-04-13
Introducing GET_MASTER_FINGERPRINT call (for legacy part of ledger-bitcoin python package)

iartemov-ledger:
And redirecting to this call all SW Wallet fingerprint calculations.

### DIFF
--- a/bitcoin_client/ledger_bitcoin/btchip/btchip.py
+++ b/bitcoin_client/ledger_bitcoin/btchip/btchip.py
@@ -56,6 +56,9 @@ class btchip:
 	BTCHIP_INS_GET_FIRMWARE_VERSION = 0xc4
 	BTCHIP_INS_COMPOSE_MOFN_ADDRESS = 0xc6
 	BTCHIP_INS_GET_POS_SEED = 0xca
+	# Introduced in app-bitcoin (legacy) v2.4.11 to provide a way to fetch
+	# the master key fingerprint after path derivation hardening applied
+	BTCHIP_INS_GET_MASTER_FINGERPRINT = 0xd0
 
 	BTCHIP_INS_EXT_GET_HALF_PUBLIC_KEY = 0x20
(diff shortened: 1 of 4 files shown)


bitcoin_client/ledger_bitcoin/btchip/btchip.py
```diff
@@ -56,14 +56,15 @@ class btchip:
 	BTCHIP_INS_GET_FIRMWARE_VERSION = 0xc4
 	BTCHIP_INS_COMPOSE_MOFN_ADDRESS = 0xc6
 	BTCHIP_INS_GET_POS_SEED = 0xca
+	BTCHIP_INS_GET_MASTER_FINGERPRINT = 0xd0
```
bigspider, line 61: might be good to add a small comment to give some context (e.g. specify what version of the app this was added in, and why), since this is otherwise a legacy file that doesn't expect changes

bitcoin_client/ledger_bitcoin/client_legacy.py
```diff
@@ -80,6 +80,20 @@ def __init__(self, comm_client: TransportClient, chain: Chain = Chain.MAIN, debu
         if self.app.getAppName() not in ["Bitcoin", "Bitcoin Legacy", "Bitcoin Test", "Bitcoin Test Legacy", "app"]:
             raise ValueError("Ledger is not in either the Bitcoin or Bitcoin Testnet app")
 
+
+    def _get_master_fingerprint(self) -> bytes:
+        """Return the 4-byte master key fingerprint.
```
bigspider: The comment in getMasterFingerprint says "Returns", this "Return".

bigspider (dismissed):
Nits

bitcoin_client/ledger_bitcoin/client_legacy.py
iartemov-ledger: Fixed.

bigspider (approved):

Issue #467, opened 2026-04-20
Avoid harcoded APDU exchange flows in Rust tests, and run test suite with the `paranoid_client` feature


bigspider:
Currently, one test fails when the `paranoid_client` feature is enabled - not noticed in CI because tests run with `--no-default-features`. This is due the hardcoded APDU flows embedded in the tests of the Rust client, which don't match the actual ones when the feature is enabled.

It would be better to avoid such hardcoded exchanges and make the tests more flexible (and run the CI with default features).

PR #468, opened 2026-04-20
Add support for `musig()` expressions in the Rust client; bump version

bigspider:
When the default `paranoid_client` feature is enabled, the Rust client doesn't work for wallet policies containing `musig()` key expressions.

This adds support by replacing such expressions with the corresponding aggregate synthetic xpub for the purposes of address computation, similarly to what the python client does.

Note: the new `test_replace_musigs` is not run in the CI, as it is part of the `paranoid_client` feature. Depends on #467 for fixing this. I manually verified that the test passes locally.

### DIFF
--- a/bitcoin_client_rs/src/bip327.rs
+++ b/bitcoin_client_rs/src/bip327.rs
@@ -152,18 +152,28 @@ mod tests {
         let cases: &[(&str, Option<&str>)] = &[
             // No musig: descriptor is returned unchanged
             (
-                "wpkh(tpubDE7NQymr4AFtewpAsWtnreyq9ghkzQBXpCZjWLFVRAvnbf7vya2eMTvT2fPapNqL8SuVvLQdbUbMfWLVDCZKnsEBqp6UK93QEzL8Ck23AwF/0/*)",
-                Some("wpkh(tpubDE7NQymr4AFtewpAsWtnreyq9ghkzQBXpCZjWLFVRAvnbf7vya2eMTvT2fPapNqL8SuVvLQdbUbMfWLVDCZKnsEBqp6UK93QEzL8Ck23AwF/0/*)"),
+                "wpkh(tpubDE7NQymr4AFtewpAsWtnreyq9ghkzQBXpCZjWLFVRAvnbf7vya2eMTvT2fPapNqL8SuVvLQdbUbMfWLVDCZKnsEBqp6UK93QEzL8Ck23AwF/<0;1>/*)",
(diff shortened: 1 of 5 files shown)


bitcoin_client_rs/src/bip327.rs
```diff
@@ -0,0 +1,195 @@
...
+    for pk in pubkeys.iter().skip(1) {
+        if pk != &pubkeys[0] {
+            return Some(*pk);
+        }
+    }
+    None
+}
+
+/// Performs BIP-327 key aggregation on a list of compressed public keys.
+pub fn key_agg(pubkeys: &[PublicKey]) -> Result<PublicKey, secp256k1::Error> {
```
iartemov-ledger, line 41: Could be `pub(crate)` as `replace_musigs()` below?

bitcoin_client_rs/src/bip327.rs
```diff
@@ -0,0 +1,195 @@
...
+        );
+    }
+    Ok(desc)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
```
iartemov-ledger, line 149: The existing tests are good, but could be strengthened with:

- Multiple musig() expressions in a single descriptor (e.g., tr(musig(a,b),pk(musig(c,d)))) — the loop logic handles this but it's untested.
- Empty musig() (0 keys) — currently would fail at xpub parsing, but an explicit test would document the behavior.
?

iartemov-ledger (commented):
2 AI inspired comments. 

I trust you on musig2 of course, but would not it be better to fix https://github.com/LedgerHQ/app-bitcoin-new/issues/467) at 1st and then to test completely this PR on CI?

bitcoin_client_rs/src/bip327.rs
bigspider: `bip327` is a private module, so its public members are still not visible from outside the crate.

bigspider:
> I trust you on musig2 of course, but would not it be better to fix #467) at 1st and then to test completely this PR on CI?

In an ideal world, yes - but #467 is not a small one, and I'm not planning to work on it very soon.

bitcoin_client_rs/src/bip327.rs
bigspider: Added both in a4504a8e01703a156ce19a70dc93e686e1db57a1

bitcoin_client_rs/src/bip327.rs
iartemov-ledger: Yes, OK, just in order to be aligned with  `replace_musigs()`.

iartemov-ledger (approved):

PR #470, opened 2026-04-22
Complete Musig2 support in the Rust client

bigspider:
#468 added support for address validation when registering wallet policies using MuSig2, but end-2-end support also requires a breaking change in the `sign_psbt` method of the client, allowing it to return pubnonces and partial signatures.

This implements a generic tag-based system for the yielded objects returned by the app during sign_psbt, similar to how it is implemented in the Python client.

The version of the client is bumped to `0.6.0`.


bitcoin_client_rs/src/protocol.rs
```diff
@@ -0,0 +1,361 @@
+//! This module contains types that are specific to the Ledger Bitcoin application protocol.
+
+use bitcoin::{
+    consensus::encode::{deserialize_partial, VarInt},
+    hashes::Hash,
```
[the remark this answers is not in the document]
bigspider: It's a trait, and it's used.

bitcoin_client_rs/src/protocol.rs
```diff
@@ -0,0 +1,361 @@
...
+#[derive(Debug, Clone)]
+pub enum SignPsbtYieldedObject {
+    /// A regular partial signature for a PSBT input.
+    Partial(PartialSignature),
+    /// A MuSig2 public nonce (round 1).
+    MusigPubNonce(MusigPubNonce),
+    /// A MuSig2 partial signature (round 2).
+    MusigPartialSignature(MusigPartialSignature),
+    /// An unknown payload with an unrecognized tag; the original bytes are included for reference.
+    Unknown(Vec<u8>),
```
iartemov-ledger, line 64: No `Unknown` in the pythopn client https://github.com/LedgerHQ/app-bitcoin-new/blob/develop/bitcoin_client/ledger_bitcoin/client_base.py#L123.

iartemov-ledger (approved):
With a minor remark.

bitcoin_client_rs/src/protocol.rs
bigspider: It could make sense to add it indeed, for the same forward-compatibility reasons it's added to the Rust client. Currently, the python client would treat an unknown 'tag' like `0xFFFFFFF8` as a the "number of inputs", per the old protocol, which is not great.

I'll add a small commit to add it there (not updating the JS client yet, as there's no MuSig2 support there, yet).

bigspider: Actually, I'll do that in a separate PR.

PR #471, opened 2026-04-23
Use default version for clang-format

bigspider:


iartemov-ledger (approved):
1 It was `12` on our side, in https://github.com/LedgerHQ/ledger-app-workflows/blob/master/.github/workflows/reusable_lint.yml#L14 it is `14`, locally I've got `16.0.3`,  whereas the latest is `22.1.4` corresponding to LLVM/Clang toolchain versioning. 
2. So no new lint warnings with the new clang-format version?

bigspider:
Yeah, it seems no new warnings between 12 and 14. Locally I have 18, which does have some differences - but at least I should be able to install 14 and align with CI!

PR #472, opened 2026-04-23
Parse unknown yielded tags in python client for sign_psbt

bigspider:
This allows future-proofing the client by returning an explicit unknown yielded value (that can safely be logged/ignored) instead of returning an incorrect enormous 'input index'.

The Rust client uses the same approach.

### DIFF
--- a/bitcoin_client/ledger_bitcoin/__init__.py
+++ b/bitcoin_client/ledger_bitcoin/__init__.py
@@ -1,7 +1,7 @@
 
 """Ledger Nano Bitcoin app client"""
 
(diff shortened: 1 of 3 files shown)


bitcoin_client/ledger_bitcoin/__init__.py
```diff
@@ -16,6 +16,7 @@
     "MusigPubNonce",
     "MusigPartialSignature",
     "SignPsbtYieldedObject",
+    "UnknownSignPsbtYieldedObject",
```
iartemov-ledger, line 19: Do we need patch version bump?

bigspider: No rush to publish the library for such a small change, it will go with the next version, whenever it's released

iartemov-ledger (approved):

PR #482, opened 2026-05-04
enable build comparison and stack usage report

mbrousset-ledger:


iartemov-ledger (approved):
What will happen if I enabled `post_stack_consumption` for tets run, but I do not enable `enable_stack_consumption` for the build?

mbrousset-ledger:
@iartemov-ledger good question, you would end up with the following error : https://github.com/LedgerHQ/ragger/pull/274/changes#diff-1b2b188c95a56abd9df951c682ae612cd878b9f89087dcbe1e68a62333158662R157

PR #483, opened 2026-05-05
Delete legacy `!IO_REVAMPED` code

bigspider:
Legacy IO is no longer used.


iartemov-ledger:
Checking with @mbrousset-ledger  the weird `nanox` only `.text` size increase, not visible locally.

mbrousset-ledger:
@iartemov-ledger it was a problem in `reusable_build.yml`, it's fixed. I re-ran the CI and delta is now 0 for all targets.

iartemov-ledger (approved):

PR #485, opened 2026-05-06
Fix breaking changes with upstream `bitcoin-core` and `rust-bitcoin`

bigspider:
This fixes two breaking changes from upstream dependencies:
- Some of the psbts generated in the test suite had `nVersion = 1`, which made `combinepsbt` fail (since changing the `nVersion` results in a different sighash). Not sure exactly why this only happens now, but forcing all PSBTs to use transaction version `2` fixes it
- `rust-bitcoin` v0.32.9 changed the meaning of `VarInt` to only support numbers up to `0x02000000`, instead of arbitrary `u64`. We implement our own version without the limitation.

This also bumps the version of the Rust client to `0.6.1`.

### DIFF
--- a/tests/test_e2e_miniscript.py
+++ b/tests/test_e2e_miniscript.py
@@ -104,6 +104,8 @@ def run_test_e2e(navigator: Navigator, client: RaggerClient, wallet_policy: Wall
 
     psbt = PSBT()
     psbt.deserialize(psbt_b64)
+    psbt.tx.nVersion = 2  # Ensure transaction version 2 (walletcreatefundedpsbt may produce version 1)
+    psbt_b64 = psbt.serialize()
 
     hww_sigs = client.sign_psbt(psbt, wallet_policy, wallet_hmac, navigator,
                                 instructions=instructions_sign_psbt,
--- a/tests/test_e2e_tapscripts.py
+++ b/tests/test_e2e_tapscripts.py
@@ -101,6 +101,8 @@ def run_test_e2e(navigator: Navigator, client: RaggerClient, wallet_policy: Wall
 
     psbt = PSBT()
     psbt.deserialize(psbt_b64)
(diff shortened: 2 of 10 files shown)


tests/test_e2e_tapscripts.py
```diff
@@ -133,6 +135,10 @@ def run_test_e2e(navigator: Navigator, client: RaggerClient, wallet_policy: Wall
         partial_psbts.append(partial_psbt_response["psbt"])
 
     # ==> finalize the psbt, extract tx and broadcast
+    print("Combining PSBTs...")
```
iartemov-ledger: Do we need this debug print?

tests/test_e2e_miniscript.py
```diff
@@ -104,6 +104,8 @@ def run_test_e2e(navigator: Navigator, client: RaggerClient, wallet_policy: Wall
 
     psbt = PSBT()
     psbt.deserialize(psbt_b64)
+    psbt.tx.nVersion = 2  # Ensure transaction version 2 (walletcreatefundedpsbt may produce version 1)
```
iartemov-ledger, line 107: It's weird, but it works on my side without the the version 2 assignment.

iartemov-ledger (dismissed):
Approved with 2 minor comments.

tests/test_e2e_tapscripts.py
bigspider: Oops, that was unintentional. Removed (with rebase).

tests/test_e2e_miniscript.py
bigspider: Did you recompile bitcoin from the current `master` branch?
I could only reproduce locally by recompiling.

iartemov-ledger: No, I did not, but OK, I see.

iartemov-ledger (approved):

PR #486, opened 2026-05-07
add CODEOWNERS

mbrousset-ledger:


iartemov-ledger (approved):
OK, very elegant and human size CODEOWNERS file. 
@bigspider , are you OK with that?

PR #488, opened 2026-05-11
Remove unnecessary port bindings in CI jobs

bigspider:
It might be the cause of intermittent CI failures observed during container setup, like [this](https://github.com/LedgerHQ/app-bitcoin-new/actions/runs/25662009806/job/75325237548).


iartemov-ledger (approved):
We dont't have  most probably anything outside containers connecting to Speculos for these tests.

PR #490, opened 2026-05-12
Fixes for baseapp

bigspider:
Fixes in this PR:
- allow signing flow to complete when the wallet policy id is passed as `NULL` (32 `0x00` bytes)
- make sure that the relevant checks in `preprocess_inputs` (witness-utxo/non-witness-utxo presence, sighash flags) are performed also on external inputs, avoiding the burden for the derived app

This makes external inputs not supported if they are:
- legacy, or nested SegWit
- SegWit version 0 with missing non-witness-utxo

Most likely, derived apps would prefer taproot, or at most segwit P2WSH, so this limitation is unlikely to be problematic.

The commits in this PR can be merged with b717a3891508fd8364a8d413b874324c1fd63655 during rebase.


3697503 (approved):

src/handler/sign_psbt.c
```diff
@@ -752,14 +753,34 @@ preprocess_inputs(dispatcher_context_t *dc,
...
-                continue;
+                // unlike the Bitcoin app, we still validate external outputs, as the derived app might sign for them
             } else {
                 bitvector_set(internal_inputs, cur_input_index, 1);
                 st->internal_inputs_total_amount += input.prevout_amount;
             }
         }
 
-        int segwit_version = get_policy_segwit_version(st->wallet_policy_map);
+        int segwit_version = -1;
```
iartemov-ledger: A trailing space here.
Also need to run the clang format as it has deviated. Not only this file, but few others:
```
#!/bin/bash

TARGET_DIR="./src"

# 1. Check if the target directory exists
if [ ! -d "$TARGET_DIR" ]; then
  echo "Error: Directory '$TARGET_DIR' does not exist."
  exit 1
fi

# 2. Check if clang-format is installed on the system
if ! command -v clang-format &> /dev/null; then
  echo "Error: 'clang-format' could not be found. Please install it first."
  exit 1
fi

echo "Formatting all .c and .h files in $TARGET_DIR..."

# 3. Find all .c and .h files and pass them to clang-format
# The {} + syntax appends all found files to a single command, which is much faster
# than running clang-format individually for every single file.
find "$TARGET_DIR" -type f \( -name "*.c" -o -name "*.h" \) -exec clang-format -i {} +

echo "Formatting complete!"
```
To think to add the clang-format CI for `baseapp` as target.

iartemov-ledger (commented):
Good change!
Only a formatting remark.

src/handler/sign_psbt.c
bigspider: I fixed the missing trailing space.

Not sure about adding clang and making processes (and rebases) more complicated, but if we do, that can be a separate PR.

iartemov-ledger: But here there is a clang deviation in your change-set in the file  ;)

bigspider: ok, passed it through clang-14 in 9fa28c8e6199af9eb40886eb6d385d0e52194838

iartemov-ledger (approved):

bigspider:
`baseapp` branch updated.

PR #494, opened 2026-05-26
Clear text representation of wallet policies

bigspider:
TODOs:

- [x] document how the `confusion_score` is computed (maybe in the Rust crate)
- [ ] currently, the pairList of `nbgl_useCaseReviewLight` is changed to allow wrapping; that's good for the cleartext description (so it only wraps on words), but it makes descriptor templates wrap unnecessarily; it would be great to not wrap the descriptor template, but it is not allowed by Nbgl with ReviewLight (all pairs must share the same wrapping behavior).
- [ ] <del>question: **should we show the address_index/change pair when verifying a receiving address?** The user should make sure that when they receive for the first time on a registered account, they receive on address index 0. Otherwise, bruteforcing from the cleartext becomes a lot harder as one would have to try address indexes to verify if money was sent there, which makes the problem a lot harder.</del> Not implemented in this release. It would be more important for a future release where the descriptor template is hidden, since the user can't verify that it matches the backup. In this release, the user is still expected to check that the backup matches, so this is not a security concern.
- [ ] <del>**simplification**: currently, the same pattern-matching machinery is used for both the top-level descriptors, and for taproot leaves. It might be much simpler to manually implement the matching for top-level descriptors (since it seems likely that it will be a fixed list that doesn't grow), and limit the pattern matching logic to the taproot leaf patterns. This doesn't require any functional change.</del>  Let's keep the full generality for now.

Enhances the UX for the registration of multisig and (taproot) miniscript wallet policies by identifying and explaining in human terms the meaning of the descriptor template.

## Framework

The cleartext representation has the following properties:
- **composable**: instead of a list of hardcoded patterns known patterns, a core set of pattern is identified that, with a limited amount of composability, aims to cover the vast majority of the descriptor templates that people are likely to use in practice, despite inherently not being as general as the full set of miniscript wallet policies.
- **reversible**: going from a descriptor template to a clear-text description is a lossy process: there could be many different descriptor templates that have essentially the same meaning. The framework guarantees that it is possible to find the original descriptor template among a reasonably sized list of options, starting from the cleartext description. (this only implemented in the Rust version).
- **extensible**: the specifications of the patterns, and a number of test cases are listed in TOML files; the C implementation and the unit tests are automatically generated with a python script (in C), or with a `build.rs` file in the reference Rust crate. As long as the language's grammar doesn't change, adding some new patterns only requires updating the TOML files, and regenerating the code.

The model introduces a `confusion_score` that allows to provide an upper bound on the number of different descriptor templates that could possibly generate the same cleartext description. For any descriptor template with a too high `confusion_score`, or with no cleartext representation for any other reason, the descriptor template is shown as usual, in order to not prevent more advanced users from benefiting from the generality of miniscript.

This guarantees that, even if the user does not have a backup of the descriptor template, but has an exact backup of the cleartext description, it is possible to programmatically enumerate all the possible descriptor templates, and eventually find the correct one. In this way, the user is protected from **ransom attacks** even if they only check that they have a backup for the cleartext, rather than the descriptor template.

**NOTE**: in this PR, the descriptor template is still always shown (except in few very simple multisigs as explained below). Therefore, _there is no ransom attack risk if the user checked that the descriptor template matches their backup_, which remains the expected workflow for the time being.

## C implementation

The implementation in C only needs:
- descriptor_template to cleartext conversion
- confusion_score computation (to make sure it is reasonable)

The full reversible process (validating that the grammar is unambiguous, and that bruteforcing a cleartext description is feasible) is implemented in [this rust crate](https://github.com/bigspider/bip388).

In this PR, the descriptor template is still shown to the user _in addition_ to the descriptor template. In the future, we could have as an opt-in feature to hide the descriptor_template (if the software wallet is prepared to show the identical clear-text to the user).

### Clear text miniscript wallet policies

For any taproot-based miniscript wallet policy that admits a cleartext description, it is shown before the descriptor template when registering the account on the device:

When the cleartext description is not available (either a pattern not covered by the specs, or a confusion score that is too high), only the descriptor template is shown (same UX as today).

Many other examples can be found in [test_vectors.toml](https://github.com/LedgerHQ/app-bitcoin-new/blob/cleartext/specs/bip388/test_vectors.toml).

### Simple multisig policies

Only for simple multisig descriptor templates, the app automatically hides the descriptor template, as the ambiguity is very low. For example, this is for a 2-of-2:

The wording is "Each of ..." for any _n_-of-_n_ multisignature (regardless if it's legacy, segwit, or even taproot musig), while it says "Any <k> of ..." for a _k_-of-_n_ multisig (with _k_ < _n_)

## Future extensions

- If software wallets also implement the same cleartext description framework, then it is possibly safe to omit the descriptor template, yet remain safe. This will have to remain an opt-in feature, in order to not disrupt existing user experiences where software wallets show the descriptor template and expects the user to verify it.
- More patterns could be added to the specifications and be supported with cleartext registration, informed by real-world usage.


unit-tests/test_cleartext.c
```diff
@@ -0,0 +1,141 @@
...
+        }
+        // Only validate the rendered line text when the descriptor renders
+        // fully to cleartext. When has_cleartext == false, at least one tap-leaf
+        // has no cleartext form and our encoder emits the fixed "(unknown)"
+        // marker, whereas the shared vectors carry the reference Rust unparser's
+        // raw-leaf rendering ("Raw policy: ..."); the text diverges by design, so
+        // we check only n_lines and the has_cleartext flag (above) for these.
+        // The app never shows the cleartext screen unless every line has
+        // cleartext, so this divergence is not user-visible.
+        if (v->cleartext_flag) {
```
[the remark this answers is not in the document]
bigspider: It's fine, the discrepancy is documented and is a limitation of the C implementation,

src/ui/display.c
```diff
@@ -139,6 +141,15 @@
     state->n_keys = wallet_header->n_keys;
     state->wallet_name = wallet_header->name;
     state->descriptor_template = descriptor_template;
+
+    LEDGER_ASSERT(n_cleartext_lines <= CT_MAX_LINES, "Too many cleartext lines");
+    if (cleartext_lines == NULL || n_cleartext_lines == 0) {
+        state->cleartext_lines = NULL;
+        state->n_cleartext_lines = 0;
+    } else {
+        state->cleartext_lines = *cleartext_lines;
```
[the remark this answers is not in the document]
bigspider: This seems to deserve a separate refactoring of the entire module, as the same concern would apply to other things copied to the `state` global.

iartemov-ledger:
> In this PR, the descriptor template is still shown to the user in addition to the descriptor template. 
in addition to the clear-text representation?

dev-tools/gen_cleartext/gen.py
```diff
@@ -0,0 +1,954 @@
...
+        else:
+            # Encode as UTF-8 byte escapes
+            for b in ch.encode("utf-8"):
+                result.append(f"\\x{b:02x}")
+        result.append("")
+    result.append('"')
+    return "".join(result)
+
+
+def main() -> int:
```
iartemov-ledger: Maybe to improve the user-facing part of the script:
```
python3 ./dev-tools/gen_cleartext/gen.py --help
wrote src/common/cleartext_specs.h
wrote src/common/cleartext_specs.c
wrote src/common/cleartext_match.c
wrote unit-tests/cleartext_vectors.inc.c
```

src/common/cleartext.c
```diff
@@ -0,0 +1,1014 @@
+/*****************************************************************************
```
iartemov-ledger, line 1: Maybe at some point we'll be able to use the `cleartext` module from Vanadium instead of re-written C-implementation?

src/common/cleartext.c
```diff
@@ -0,0 +1,1014 @@
...
+ *   Ledger App Bitcoin.
+ *   (c) 2026 Ledger SAS.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *****************************************************************************/
+
+#include <stdint.h>
+#include <stdbool.h>
```
iartemov-ledger, line 10: Got a vibe-coded "BIP388 descriptor-template -> cleartext encoder" python script based on `specs/bip388/cleartext.toml`:
```
python3 dev-tools/gen_cleartext/cleartext_py.py "wsh(sortedmulti(2,@0/**,@1/**))"
{
  "cleartext": [
    "Each of @0 and @1 must sign"
  ],
  "has_cleartext": true,
  "confusion_score": 7
}
```

Could be useful ?

iartemov-ledger (commented):
Started looking at and playing with. 


Few Claude's comments I will need to decode yet:

 # `cleartext` Implementation Review

Review of the cleartext wallet-policy feature on the `cleartext` branch of
`app-bitcoin-new` (the C base app), compared against the Rust reference
implementation in `vanadium/apps/bitcoin/bip388`.

Findings are ranked from **most severe** to **least severe**.

---

## 1. Tap-leaf ordering can diverge from the reference canonical order
**Severity: high — correctness, user-visible**

`leaf_cmp` (`src/common/cleartext.c:471-520`) compares bindings positionally in
pattern order. For a `Multisig` / `SortedMultisig` leaf the bindings are
`[threshold, keys]`, so C sorts by **threshold first, then key count**. The
reference `display_cmp`
(`vanadium/apps/bitcoin/bip388/src/cleartext/mod.rs:194-197`) sorts by
**key count first, then threshold**:

```rust
k1.len().cmp(&k2.len()).then(t1.cmp(t2)).then_with(|| cmp_keys(k1,k2))
```

These disagree whenever two same-class multisig leaves have threshold order and
key-count order inverted. For
`tr(@0/**,{multi_a(3,@1..@4),multi_a(2,@5..@9)})` the reference produces
(verified):

```
Any 3 of @1, @2, @3 and @4 must sign
Any 2 of @5, @6, @7, @8 and @9 must sign
```

but C's `leaf_cmp` emits the 2-of-5 first (threshold `2 < 3`). This is a
*fully-cleartext* taproot, so it is user-visible, and it breaks the
canonical-ordering assumption the Rust reverse-decoder relies on. No current
vector exercises it, so the unit test passes.

**Fix:** for the multisig classes, compare key count → threshold → keys
explicitly rather than relying on positional binding order. Add a regression
vector.

---

## 2. No CI guard that generated C matches the TOML
**Severity: high — maintainability, can ship wrong code silently**

`cleartext_match.c`, `cleartext_specs.c`, and
`unit-tests/cleartext_vectors.inc.c` are committed generated artifacts, but
nothing fails the build if someone edits `specs/bip388/cleartext.toml` without
rerunning `gen.py`, or hand-edits the generated C.

**Fix:** a one-line CI step — run `gen.py`, then `git diff --exit-code`. Highest
value-to-cost ratio of anything here.

---

## 3. `(unknown)` leaves weaken the stated guarantee
**Severity: high — security-relevant; hard blocker for the opt-in feature**

`cleartext_encode` (`src/common/cleartext.c:989-999`) renders every
unrecognized leaf as the fixed `"(unknown)"`, where the reference prints
`"Raw policy: <descriptor>"`. The framework's premise is that a cleartext
backup suffices to recover the template; `(unknown)` does not — two different
unknown leaves are indistinguishable.

It is safe *today* only because such policies always have
`has_cleartext == false`, so the descriptor template is always shown and the
cleartext screen is suppressed. But:

- the C output is **untestable** against the shared vectors for these cases
  (`unit-tests/test_cleartext.c:119` skips text comparison when
  `!cleartext_flag`) — that path has zero text coverage;
- it is a landmine for the planned "hide descriptor template" opt-in.

**Fix:** add even a minimal AST unparser (multisig / `pk` only), or at minimum
document this as a hard prerequisite for the opt-in.

---

## 4. Two diverging copies of the spec
**Severity: medium — drift risk**

This repo's `specs/bip388/cleartext.toml` + `test_vectors.toml` are independent
from vanadium's. They are the "single source of truth" for two implementations
that can silently drift apart.

**Fix:** vendor one canonically (submodule or sync script), and/or a CI diff
between the two trees.

---

## 5. `leaf_cmp`'s `CT_BV_KEY` musig branch is a silent no-op
**Severity: medium — latent trap**

`src/common/cleartext.c:481-485` only tie-breaks when both keys are
`KEY_EXPRESSION_NORMAL`; for two musig keys they fall through comparing nothing
(reference `cmp_key` compares len then indices). Dead today (musig leaves
classify as `Multisig`, so `$key` only ever binds a plain key), but a trap if a
future spec binds a musig to a `$key`.

**Fix:** assert plainness, or implement the musig comparison.

---

## 6. Policy is classified ~2× and tap-leaves matched many times
**Severity: low/medium — efficiency, on-device**

`src/handler/register_wallet.c:229-243` calls `cleartext_confusion_score` then
`cleartext_encode`; both run `match_top_level`, collect leaves, and
`match_tapleaf` per leaf, and both run `key_orderings_count` (encode
`(void)`-discards the orderings it computes at `src/common/cleartext.c:916`).
Inside encode, `match_tapleaf` is re-invoked on the same sub-nodes by `leaf_cmp`
(`:503-504`), `render_spec_at` (`:779`), and `leaf_score` (`:838`).

**Fix:** classify once into a small cached structure and thread it through
score + encode + sort + render. (Minor on small policies; real for large/deep
taptrees.)

---

## 7. `CT_MAX_LINES == 8` silently caps describable taproots
**Severity: low — product limit**

`src/common/cleartext.h:15`: a taproot with >7 recognized leaves falls back to
raw even though it is fully describable. Probably fine, but worth a comment at
the cap so it is an intentional decision, not a surprise.

---

## 8. `CT_MAX_TAP_LEAVES 64` vs. comment saying depth-9 allows up to 256
**Severity: low — comment/constant mismatch**

`src/common/cleartext.c:807`: the cap is a safe-fail (returns `UINT64_MAX` →
treated as too-confusing), but the mismatch between comment and constant invites
confusion.

---

## 9. Duplicated threshold/keys scan
**Severity: low — minor cleanup**

`admitting_pattern_count` (`src/common/cleartext.c:89-108`) and
`bindings_is_n_of_n` (`:719-728`) run the same scan loop. Extract one helper.

---

### Suggested first actions

1. **#1** — ordering fix + regression vector.
2. **#2** — CI regen guard.

These are the two highest-impact items.


Finally first few seem to be relevant, are not they ?

src/common/cleartext.c
bigspider: My first thought is that it's not possible because they don't share the same structs (e.g. `policy_node_t` etc. used in the bitcoin app). But indeed, recompiling adds a lot of complexity.

A clear binding from Rust would probably need to take the descriptor template as a string, and return the cleartext as a list of strings. Therefore, it would have to parse the descriptor template into Rust's version of the AST.

Performance is probably not an issue, though, and avoiding the rewrite would be a big win if the increase in code size is not insane.
I'll try to see if I can make usable C bindings from the Rust binaries. It might still be a better approach, everything considered!

src/common/cleartext.c
bigspider: Similarly to your [comment above](https://github.com/LedgerHQ/app-bitcoin-new/pull/494#discussion_r3453631754), for python and other high-level languages, bindings from Rust might be more maintainable.

src/common/cleartext.c
bigspider: I did some attempts with Claude at
- modifying the Rust crate to be able to generate the C bindings ([attempt1](https://github.com/LedgerHQ/vanadium/pull/274), [attempt2](https://github.com/LedgerHQ/vanadium/pull/275))
- integrating it with the bitcoin app ([attempt1](https://github.com/LedgerHQ/app-bitcoin-new/pull/509), [attempt2](https://github.com/LedgerHQ/app-bitcoin-new/pull/510))

In both cases, the binary size increases by unacceptable amounts (~36kb in the second, best attempt).

Unfortunately, while there is some room for improvement, the analysis (with the help of Claude) is that the binary size would increase by ~24 kb _at least_, because the Rust crate and the C app have different AST (and parsing code) to represent descriptor templates, and the size of that code is already in the 18kb range.

Therefore the approach seems not viable.

dev-tools/gen_cleartext/gen.py
bigspider: Improved the script to support `--help` and `--check` to verify instead of generating in aa9658bcbdae1de8d496f5b998ec15b418269b89.

Added a call to `--check` to the CI in 7b80557b0ba7b0e6ee6d3e6cf5aa92c347cc1e24.

bigspider:
Going through the comments [from above](https://github.com/LedgerHQ/app-bitcoin-new/pull/494#pullrequestreview-4544837101):

> ## 1. Tap-leaf ordering can diverge from the reference canonical order

Added test cases upstream and fixed the canonical ordering in eb387e471bba3acfd1545a4feb2685b2914e9353.
 
> ## 2. No CI guard that generated C matches the TOML

Addressed with 7b80557b0ba7b0e6ee6d3e6cf5aa92c347cc1e24.

> ## 3. `(unknown)` leaves weaken the stated guarantee

It's a fair point, but it's not easy to fix without an AST unparser (that is, producing the descriptor template from the parsed AST), which is currently missing and non-trivial to build. Since we don't show the cleartext unless all leaves have the cleartext, this is not a problem.

> ## 4. Two diverging copies of the spec

Fair point, but this can only be addressed once the upstream Rust crate has a permanent house, possibly separate from the Vanadium repository.

> ## 5. `leaf_cmp`'s `CT_BV_KEY` musig branch is a silent no-op

Done in ee5f859c74895a226aa86d9f402720cd42966b5e.

> ## 6. Policy is classified ~2× and tap-leaves matched many times
I'm not worried about performance for this part of the code.

> ## 7. `CT_MAX_LINES == 8` silently caps describable taproots

That's fine, it's an arbitrary limitation by design.

> ## 8. `CT_MAX_TAP_LEAVES 64` vs. comment saying depth-9 allows up to 256

I don't see the problem, and 64 is plenty in practice, unlikely we'll ever need more.

> 
> ## 9. Duplicated threshold/keys scan

Not worth the refactor imho.

specs/bip388/test_vectors.toml
```diff
@@ -0,0 +1,884 @@
+# Test vectors for BIP388 cleartext display.
```
iartemov-ledger, line 1: Should we add test vectors with a finite, correctly-computed score > 100 000:
```
- tr(@0/**,{{{pk(@1/**),pk(@2/**)},{pk(@3/**),pk(@4/**)}},{{pk(@5/**),pk(@6/**)},{pk(@7/**),pk(@8/**)}}})
- tr(@0/**,{{{pk(@1/<0;1>/*),pk(@1/<2;3>/*)},{pk(@1/<4;5>/*),pk(@1/<6;7>/*)}},{pk(@1/<8;9>/*),pk(@1/<10;11>/*)}})
```

And maybe even a functional test?

src/handler/register_wallet.c
```diff
@@ -223,10 +224,46 @@ void handler_register_wallet(dispatcher_context_t *dc, uint8_t protocol_version)
...
+            n_cleartext_lines = 0;
+        }
+    }
+
+    // For multisig policies (all coalesced into DC_MULTISIG),
+    // the cleartext "Any K of <keys> must sign" captures the spending policy
+    // with little risk of ambiguity. Therefore, it is safe to hide the
+    // raw descriptor template, simplifying the UX.
+    const char *descriptor_to_show = (char *) policy_map_descriptor;
+    if (has_cleartext && n_cleartext_lines > 0 && cleartext_class == DC_MULTISIG) {
```
iartemov-ledger, line 257: - we might want to use another, small  threshold as additional factor to decide to hide the descriptor template? 
- apparently we hide the address type / script encoding and the sorted-ness here (for `DC_MULTISIG`). But it's the decision made here
- should we be able to turn on/off the mechanism through menu for example?

iartemov-ledger (commented):
Just a couple of comments. 
Plus attaching generated documents to facilitate my understanding:
- [register_wallet_cleartext_flex_sequences.html](https://github.com/user-attachments/files/30078216/register_wallet_cleartext_flex_sequences.html)
- 
wallet_policy_classification

src/handler/register_wallet.c
bigspider: > * we might want to use another, small  threshold as additional factor to decide to hide the descriptor template?

I considered this, but I think it's not worth the complexity; plus, for non-multisig miniscript policies, people are expecting to see the descriptor and compare with what their software wallet is saying - so hiding it would be surprising.
For a future version, as we discussed, hiding the descriptor template could be enabled with explicit opt-in from the software wallet (who could show the same cleartext to the user, instead of the descriptor.

> * apparently we hide the address type / script encoding and the sorted-ness here (for `DC_MULTISIG`). But it's the decision made here

Yeah, that's on purpose.

> * should we be able to turn on/off the mechanism through menu for example?

I don't think so - since we only do it for simple multisig, there is really no added risk for the user.

iartemov-ledger:
2 test cases more if you're interested:
- `test_register_wallet_accept_wit_2of3`
- `test_register_wallet_tr_script_multi_leaf`

[test_register_wallet.py.txt](https://github.com/user-attachments/files/30088839/test_register_wallet.py.txt)

specs/bip388/test_vectors.toml
bigspider: Good point!
Added the test for the first descriptor template above in 66f530e7083b7391db91a254d487a7310a3ab45a.

The commit also include some small refactors on the unit tests file.

bigspider:
> 2 test cases more if you're interested:
> 
> * `test_register_wallet_accept_wit_2of3`
> * `test_register_wallet_tr_script_multi_leaf`

Thanks! Added in 1307b4c2555544998435c011f5e1cb149333708b, including the snapshots.

At some point it would be good to have a separate place for generating snapshots so it doesn't mix with the _actual_ functional tests, but that's for future work.

iartemov-ledger:
> > 2 test cases more if you're interested:
> > 
> > * `test_register_wallet_accept_wit_2of3`
> > * `test_register_wallet_tr_script_multi_leaf`
> 
> Thanks! Added in [1307b4c](https://github.com/LedgerHQ/app-bitcoin/commit/1307b4c2555544998435c011f5e1cb149333708b), including the snapshots.
> 
> At some point it would be good to have a separate place for generating snapshots so it doesn't mix with the _actual_ functional tests, but that's for future work.

Do you use the snapshot generation using the CI: https://github.com/LedgerHQ/app-bitcoin/actions/workflows/build_and_functional_tests.yml -> Run workflow ->Use workflow from=your_branch, CI behavior if the test snapshots are different than expected = Open a PR?
It generates all snapshots for all devices targeting your branch and then you merge it as a separate commit without mixing with .py files.

iartemov-ledger (approved):
Great job!

bigspider:
> Do you use the snapshot generation using the CI: https://github.com/LedgerHQ/app-bitcoin/actions/workflows/build_and_functional_tests.yml -> Run workflow ->Use workflow from=your_branch, CI behavior if the test snapshots are different than expected = Open a PR? It generates all snapshots for all devices targeting your branch and then you merge it as a separate commit without mixing with .py files.

Some times I use it, but when it's not a lot of snapshots, I often still prefer to regenerate locally :)

PR #495, opened 2026-05-27
Playground CLI and TUI to interact with the app

bigspider:
Improves the txmaker with minimal support for MuSig2 PSBT fields (and addresses few other TODOs).

Adds a python playground with a CLI and a TUI to test the various (testnet) app features against speculos. A list of presets (easy to extend) allows to programmatically generate known situations, without hardcoding fixed test-cases that would only work with a fixed seed.

The code of the playground tool is of the vibecoded quality and wasn't reviewed. Only meant as a developer tool for convenience. It is all contained in the `/dev-tools/playground` folder, and has no impact on other code in the repo.

### DIFF
--- a/dev-tools/playground/README.md
+++ b/dev-tools/playground/README.md
@@ -1,0 +1,171 @@
+# Bitcoin app playground
+
+A small, dev-only playground for exercising the Ledger Bitcoin app's APDU
+commands against speculos or a real device. Each command maps to one APDU
+so you can watch the corresponding UX flow on the device screen.
+
+Two front-ends:
+
+- **CLI** ([cli.py](cli.py)) — single-shot subcommands, scriptable. Takes a
+  fully-resolved wallet policy via inline `--template` / `--key` / `--name`
+  flags; no preset concept.
+- **TUI** ([gui.py](gui.py)) — Textual-based, tabbed UI; one tab per
+  command. Adds *presets* on top: picking a preset prefills the tab's
+  fields (and, for wallet policies, eagerly resolves device-derived xpubs).
+  Streams the same stdout/stderr the CLI would print into an output panel.
+
+## Setup
+
+All commands assume you're at the repo root (`app-bitcoin-new/`).
+
+```bash
+# 1. Create and activate a Python virtualenv (Python >= 3.8).
+python3 -m venv .venv
+source .venv/bin/activate
+
+# 2. Install the ledger_bitcoin client library (editable install).
+pip install -e ./bitcoin_client
+
+# 3. Install the playground's extra dependencies.
+pip install -r dev-tools/playground/requirements.txt
+```
+
+If you also plan to use the speculos emulator, install it too:
+
+```bash
+pip install speculos
+```
+
+## Running speculos (optional, for emulator use)
+
+In a separate terminal:
+
+```bash
+speculos build/nanos2/bin/app.elf
+```
+
+(replace `nanos2` with the device target you have an SDK build for).
+
+## Running the TUI
+
+```bash
+python dev-tools/playground/gui.py
+```
+
+The TUI has one tab per command (`get-fingerprint`, `get-xpub`,
+`register-wallet`, `get-address`, `sign-psbt`, `sign-message`). At
+startup it probes speculos first, then a USB device; pass `--target
+speculos` or `--target hid` to skip the probe. The Output panel at the
+bottom streams stdout/stderr from each command in real time — so what
+you see is exactly what the CLI would have printed.
+
+### Presets
+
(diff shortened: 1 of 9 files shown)


dev-tools/playground/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Sign-PSBT input
+
+A single textarea: leave empty to generate a fake PSBT (uses the
+Inputs/Outputs fields); paste a base64 PSBT (auto-detected by the `cHNidP`
+prefix); or type a file path to load from disk.
+
+### Keyboard / mouse
+
+- `Ctrl+L` clears the log; `Ctrl+R` reconnects; `Ctrl+Q` quits.
```
iartemov-ledger: It would be useful to be able to copy logs.

dev-tools/playground/presets.py
```diff
@@ -0,0 +1,312 @@
...
+def _huge_fee_mutator(psbt: PSBT) -> PSBT:
+    """Reduce every output value to dust so that almost the entire input
+    value becomes fee — exercises the device's high-fee warning UX."""
+    DUST = 1000
+    for vout in psbt.tx.vout:
+        vout.nValue = DUST
+    return psbt
+
+
+def _zero_outputs_mutator(psbt: PSBT) -> PSBT:
```
iartemov-ledger: `_zero_outputs_mutator` is defined but never registered in SIGN_PSBT_SCENARIO_PRESETS, while the `dev-tools/playground/README.md#L81` advertises a zero-outputs-tr preset that doesn't exist.

test_utils/wallet_policy.py
```diff
@@ -36,13 +36,31 @@ class PlainKeyPlaceholder:
...
+    num2: int
+
+
+KeyPlaceholder = Union[PlainKeyPlaceholder, MuSig2KeyPlaceholder]
 
 
 def parse_placeholder(placeholder_str: str) -> KeyPlaceholder:
     """Parses a placeholder string to create a KeyPlaceholder object."""
-    if placeholder_str.startswith('@'):
+    if placeholder_str.startswith('musig('):
```
iartemov-ledger, line 55: We've got now 2 musig parsers to maintain?
The 1st one is in `test_utils/musig2.py`, `parse_placeholder()`.

iartemov-ledger (commented):
Very nice tool!
I guess we'll be able to improve it as we use it. 
Just a few remarks.

dev-tools/playground/README.md
bigspider: You can always copy individual parts (you might have to hold SHIFT to be able to select or right click - the exact ways might vary based on the terminal).

Anyway, added it as a proper feature in be0e748d591876b9165e2bc15ecd5c561e5f525b, might still be useful to copy the full logs instead.

dev-tools/playground/presets.py
bigspider: Fixed in 15c7fd76559106acfcdb3d22e3cd84312ec15eb7 - it was from a previous preset I deleted.

test_utils/wallet_policy.py
bigspider: Good catch, and in fact there were also some other duplicated types. Fixed in 15c7fd76559106acfcdb3d22e3cd84312ec15eb7.

dev-tools/playground/README.md
iartemov-ledger: Even after `xclip` installation the copy of the entire log did not work. 
It started working only with this fix:
```
diff --git a/dev-tools/playground/gui.py b/dev-tools/playground/gui.py
index 1f32de21..1b985a41 100644
--- a/dev-tools/playground/gui.py
+++ b/dev-tools/playground/gui.py
@@ -194,8 +194,13 @@ def _write_clipboard(text: str) -> bool:
     ]
     for cmd, data in candidates:
         try:
+            # Do NOT capture stdout/stderr: xclip/xsel fork a background
+            # process that keeps owning the X selection (and inherits the
+            # pipes), so capturing would block until the timeout. Redirect
+            # to DEVNULL so the call returns as soon as the parent forks.
             result = subprocess.run(
-                cmd, input=data, capture_output=True, text=True, timeout=2
+                cmd, input=data, text=True, timeout=2,
+                stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL,
             )
             if result.returncode == 0:
                 return True
```

And indeed, the partial copy works for me (Ubuntu 22.04) with a `Shift`.

bigspider: Couldn't get it to work reliably on my end, so I removed the commit with the feature.

iartemov-ledger (approved):

PR #498, opened 2026-06-03
Fix and simplify HAVE_SEMIHOSTED_PRINTF

bigspider:
It has been broken for a while, and it is still occasionally useful when the standard PRINTF fails.


iartemov-ledger (approved):
I do not use often this feature but it does work.
It gives:
```
standalone_app_main
```
instead of:
```
seproxyhal: printf: standalone_app_main
```
in Speculos log at app start.

PR #500, opened 2026-06-08
docs: Add user documentation

### LINKED ISSUE #330: Document the signing behavior from version 2.4.0
Since version 2.4.0, the app has a fine-grained signing behavior, allowing to avoid signing unnecessarily for all possible spending paths if only some are present. This should be better documented.

For MuSig2, the app's decisions are affected by the presence of pubnonces, and this might not match the expectations of people working with the PSBT signing flow of musig (particularly, the app assumes Round 2 if _any_ pubnonce is present in the PSBT.

TBD if we also want to modify the client libraries to handle this more gracefully (e.g.: if some pubnonces are present, but not all - then strip all of them, so the app proceeds to Round 1 instead of returning an error later).

bigspider:
Existing documentation is rather detailed on the wire protocol, but lacks a higher level technical overview of the features and operating principles of the app.

This commit adds two new high-level guides for technical readers that are not interested in the details of the wire protocol:

- features.md: high-level overview of the app's features
- integration.md: integration concepts and security model

Also, wallet.md was trimmed to avoid repeating what is in BIP-388.

Closes: #330

### DIFF
--- a/doc/integration.md
+++ b/doc/integration.md
@@ -1,0 +1,169 @@
+# Integration concepts and security model
+
+This document explains the concepts an integrator needs in order to build on top of the
+Ledger Bitcoin app using one of the existing client libraries
+([Python](../bitcoin_client), [JavaScript](../bitcoin_client_js),
+[Rust](../bitcoin_client_rs)). It complements the feature overview in
+[features.md](features.md) and the protocol-level specifications in
+[bitcoin.md](bitcoin.md) and [wallet.md](wallet.md).
+
+## The device is stateless
+
+The app does **not** persist any wallet-related or transaction-related information between
+calls. There is no notion of "the wallet currently loaded on the device": every request that
+involves an account (deriving an address, signing a transaction) must carry the full
+description of that account, supplied by the host.
+
+This means the **host application is responsible for storing** the account definitions
+(and, for registered wallets, the authorization HMAC described below) and providing them
+on each call. The host is *not* trusted, however: it merely supplies the information the
+device needs. Anything that matters for security is verified by the user on the device's
+trusted screen (see [On-device verification](#on-device-verification-and-the-trusted-screen)
+below), so a compromised host cannot make the device sign or display something the user
+did not approve.
+
+## Wallet policies
+
+An account is described by a **wallet policy**, which is the standard defined in
+[BIP-388](https://github.com/bitcoin/bips/blob/master/bip-0388.mediawiki). A wallet
+policy has two parts:
+
+1. A **descriptor template** — an output descriptor in which each key is replaced by a
+   key placeholder `@0`, `@1`, … The placeholder is followed by `/**` (a shorthand for
+   the receive/change derivation `/<0;1>/*`).
+2. A **key information vector** — the actual keys (`xpub`s), each optionally preceded by
+   its **key origin information**: the master key fingerprint and the derivation steps
+   used to reach that `xpub`, e.g. `[f5acc2fd/48'/1'/0'/2']xpub6E...`.
+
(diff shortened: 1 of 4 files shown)


doc/integration.md
```diff
@@ -0,0 +1,169 @@
...
+```
+wsh(sortedmulti(2,@0/**,@1/**,@2/**))
+keys: [
+  "[f5acc2fd/48'/0'/0'/2']xpub6E...",   // this device
+  "[1a2b3c4d/48'/0'/0'/2']xpub6E...",   // cosigner 2
+  "[5e6f7a8b/48'/0'/0'/2']xpub6E..."    // cosigner 3
+]
+```
+
+A miniscript "recovery vault" — spendable by the primary key at any time, or by a backup
```
iartemov-ledger: Should also precise " (must be registered before use)" at the end.

doc/wallet.md
```diff
@@ -1,191 +1,106 @@
...
 
-The `/**` in the placeholder template represents commonly used paths for
-receive/change addresses, and is equivalent to `<0;1>`.
+For a higher-level explanation of wallet policies, registration, and the security model,
+see [integration.md](integration.md).
 
-The placeholder `@i` for some number *i* represents the *i*-th key in the
-vector of key origin information (which must be of size at least *i* + 1,
-or the wallet policy is invalid).
+## Supported scripts
```
iartemov-ledger, line 21: There is a bit of redundancy between this document/part and:
- https://github.com/LedgerHQ/app-bitcoin-new/blob/docs2/doc/features.md#supported-account-and-script-types
- https://github.com/LedgerHQ/app-bitcoin-new/blob/docs2/doc/integration.md#wallet-policies
Would not it be better to concentrate all the info in `wallet.md` and from `features.md` and `integration.md` to make only references?

doc/integration.md
bigspider: I didn't like repeating that, so I moved it instead into its own separate line in 6756ab29358e640f4721ca1cb71183dd69024af5, before both examples that have to be registered.

doc/wallet.md
bigspider: Good point, I changed that in 6756ab29358e640f4721ca1cb71183dd69024af5, simplifying `features.md` and making `wallet.md` the source of truth for these part of the specs.

iartemov-ledger: It seems that there is still a redundancy:
-  Registration / HMAC mechanism - duplicated nearly verbatim:
    - [wallet.md:96-103] "Registration and usage": stateless -> returns 32-byte HMAC-SHA256, key derived from seed via SLIP-0021, "registration is non-revocable."
    - [integration.md:94-110]"Registration and the HMAC": same content - 32-byte HMAC-SHA256, SLIP-0021 seed derivation, "registration is non-revocable."

- Wallet-policy -  defined twice (both give the full two-part definition with key-origin-information explanation):
    - [wallet.md:3-14]
    - [integration.md:25-43] 


The judgment from Claude:


> If you want full consistency with the source-of-truth principle, the natural split is: wallet.md owns the spec (definition, serialization, HMAC mechanics, SLIP-0021, bounds), and integration.md keeps only the conceptual "why" (why registration protects against a malicious host, the security property) and references wallet.md for the HMAC/serialization mechanics — the same treatment features.md just received. Right now integration.md re-specifies the HMAC details rather than referencing them.

bigspider: Further simplified integration.md in 909f6a9d4b6800df5661c1a91e18e097d3adba3b.

iartemov-ledger (approved):
Cool! You might only want to squash the commits ;)

PR #502, opened 2026-06-10
Add Absolution-based fuzzing

N3ur0sis:
## Description
Integrates the Secure SDK's Absolution-based  fuzzing framework into the
Bitcoin app. The app opts in with a small `fuzzing/` subtree and
`cmake/LedgerAppFuzz.cmake`
The fuzzer input is split into `[ prefix | tail ]`: the prefix restores the
app's globals to a target state from a declarative model, and the tail is fed to
the real APDU dispatcher. This reaches deep, multi-APDU PSBT/wallet logic
(merkleized PSBTs, wallet-policy registration, swap checks) at near-stateless
speed.
## Changes
- `fuzzing/` — app harness, semantic host/model layer, mocks, invariants, and a
  checked-in base corpus.
- `.clusterfuzzlite/` + `.github/workflows/clusterfuzzlite.yml` — ClusterFuzzLite CI.
- Small defensive fixes found by the framework, each in its own commit:
  - `common/merkle`: bound `ceil_lg` and use unsigned shifts.
  - `common/parser_ext`: guard zero-length `memmove`.
- Fuzz-only relaxations in `swap_checks` / `policy` are gated behind
  `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` — no change to production builds.
Opt-in, additive tooling: no on-device/firmware behavior change. Requires the
SDK's `aro/fuzzing-framework` branch (pinned in `.clusterfuzzlite/Dockerfile`).


src/handler/lib/policy.c
```diff
@@ -2015,8 +2015,13 @@ int is_policy_sane(dispatcher_context_t *dispatcher_context,
             if (memcmp(pubkey_i.compressed_pubkey,
                        pubkey_j.compressed_pubkey,
                        sizeof(pubkey_i.compressed_pubkey)) == 0) {
+#ifndef HAVE_AUTOAPPROVE_FOR_PERF_TESTS
```
bigspider: Should this use `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`? Same for various other parts of this PR.

N3ur0sis: Yes! That makes it much more explicit for the fuzz-only relaxations. Moved them to `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` (kept HAVE_AUTOAPPROVE_FOR_PERF_TESTS just for the UI auto-approve)

.clusterfuzzlite/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM gcr.io/oss-fuzz-base/base-builder:v1
+
+# Pillow is required by the SDK glyph generator (icon2glyph.py).
+RUN pip3 install --break-system-packages --no-cache-dir pillow>=3.4.0
+
+RUN apt update && apt install -y ninja-build zip libbsd-dev pkg-config git
+
+ARG LEDGER_SECURE_SDK_REPO=https://github.com/LedgerHQ/ledger-secure-sdk.git
+ARG LEDGER_SECURE_SDK_REF=e007cf677f1f7efd0048ad60e826f97e7f173277
```
iartemov-ledger: So we're waiting until https://github.com/LedgerHQ/ledger-secure-sdk/pull/158 is merged to "unfix" the SDK ref ?

src/common/merkle.c
```diff
@@ -55,14 +55,19 @@ int merkle_get_ith_direction(size_t size, size_t index, size_t i) {
     if (size <= 1 || index >= size) {
         return -1;
     }
-
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
```
iartemov-ledger, line 58: I've got an impression that the commits' composition can be improved:

- [fuzzing: add Absolution-based stateful fuzzing framework](https://github.com/LedgerHQ/app-bitcoin-new/pull/502/commits/6ef96fbd61331743b4094f711583d59fcf06a0e0) :green_circle: 

- [fuzzing: adapt Bitcoin src for the Absolution harness](https://github.com/LedgerHQ/app-bitcoin-new/pull/502/commits/d7456f0af7563d0a622401366748e4ed8c6131be)  :red_circle:  
    => to be squashed to  [fuzzing: gate fuzz-only relaxations on FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION]

- [common/merkle: bound ceil_lg and use unsigned shifts](https://github.com/LedgerHQ/app-bitcoin-new/pull/502/commits/06aa5f83bdf963967c0c588db04c772fe93096c7)  :red_circle:  
    => the "gate fuzz-only relaxations" mixed up with the fixes - to be split, to name the fix one with "fix" prefix? 

- [common/parser_ext: guard zero-length memmove](https://github.com/LedgerHQ/app-bitcoin-new/pull/502/commits/e980b072ffb9c1b90b3e5ae0bebe6b8cdae9eb80)  :red_circle:  
    => if it is a fix then to name the fix one with "fix" prefix? 

- [fuzzing: gate fuzz-only relaxations on FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION](https://github.com/LedgerHQ/app-bitcoin-new/pull/502/commits/65657f0b47311957893788542d8c2a10b60148e6) :green_circle:

iartemov-ledger (commented):
Cool work!
1st round of the review, I played a bit with the fuzzer locally:

# Bitcoin fuzzing — coverage & UBSan cleanup

## 1. What was run

- Edited `fuzzing/harness/fuzz_dispatcher.c`: traded one `SIGN_PSBT` structured-lane
  slot for a `GET_EXTENDED_PUBKEY` slot (still 16 slots, `% 16` routing unchanged).
```bash
diff --git a/fuzzing/harness/fuzz_dispatcher.c b/fuzzing/harness/fuzz_dispatcher.c
index 2fb48fd7..54866214 100644
--- a/fuzzing/harness/fuzz_dispatcher.c
+++ b/fuzzing/harness/fuzz_dispatcher.c
@@ -86,7 +86,7 @@ static const size_t btc_raw_n_commands =

 /* Structured-lane command map (16 slots): slot count == weight. */
 const fuzz_command_spec_t fuzz_commands[16] = {
-    /* SIGN_PSBT x 8 (50%) */
+    /* SIGN_PSBT x 7 (43.75%) */
     {.cla = CLA_APP,
      .ins = SIGN_PSBT,
      .p2_max = CURRENT_PROTOCOL_VERSION,
@@ -115,8 +115,10 @@ const fuzz_command_spec_t fuzz_commands[16] = {
      .ins = SIGN_PSBT,
      .p2_max = CURRENT_PROTOCOL_VERSION,
      .flags = FUZZ_CMD_HAS_DATA},
+    /* GET_EXTENDED_PUBKEY x 1 (6.25%) — also picked in the raw lane; a
+     * structured slot drives the display/derivation path more often. */
     {.cla = CLA_APP,
-     .ins = SIGN_PSBT,
+     .ins = GET_EXTENDED_PUBKEY,
      .p2_max = CURRENT_PROTOCOL_VERSION,
      .flags = FUZZ_CMD_HAS_DATA},
     /* REGISTER_WALLET x 2 (12.5%) */
```
- Regenerated seeds: `generate-seed-corpus.py` → `fuzzing/base-corpus/`
  (needs `BOLOS_SDK` + `APP_DIR`).
- Re-ran the campaign with a real budget: `WARMUP_SEC=120 MAIN_SEC=1800 WORKERS=8`
  (was the 30 s / 60 s / 2-worker smoke test).

### Reproduce

```bash
# --- host: build the CFL image and enter it with the live tree mounted ---
# inside app-bitcoin-new folder
docker build -t btc-fuzz -f .clusterfuzzlite/Dockerfile .
docker run --rm -it \
    -v <PATH_TO_BTC_APP>:/app \
    -v <PATH_TO_SDK>:/ledger-secure-sdk \
    btc-fuzz bash

# --- inside the container ---
export BOLOS_SDK=/ledger-secure-sdk
export APP_DIR=/app
unset LIB_FUZZING_ENGINE             # else cmake links the missing /usr/lib/libFuzzingEngine.a

python3 /app/fuzzing/scripts/generate-seed-corpus.py /app/fuzzing/base-corpus
rm -rf /app/build/fast /app/build/cov /app/.fuzz-artifacts/cov-run
WARMUP_SEC=120 MAIN_SEC=1800 WORKERS=$(nproc) \
  "$BOLOS_SDK"/fuzzing/scripts/app-campaign.sh --app-dir /app cov-run

# --- inspect (artifacts appear on host via the mount) ---
cat /app/.fuzz-artifacts/cov-run/key-files-coverage.txt
grep -rhc "runtime error" /app/.fuzz-artifacts/cov-run/targets/fuzz_app/*/worker-*/fuzz.log
```

## 2. Coverage improvement (`bitcoin-run` → `cov-run`)

| Metric | Before | After | Δ |
|---|---|---|---|
| Lines | 61.34% | **76.21%** | +14.9 |
| Regions | 53.19% | 66.21% | +13.0 |
| Functions executed | 82.10% | 90.74% | +8.6 |
| Branches | 52.86% | 66.56% | +13.7 |

Biggest per-file gains (lines):

| File | Before → After |
|---|---|
| `get_extended_pubkey.c` | 20.0% → 77.4% (harness slot) |
| `stream_merkleized_map_value.c` | 0.0% → 83.3% (was reachable after all) |
| `common/script.c` | 24.3% → 65.1% |
| `sign_psbt/txhashes.c` | 38.6% → 72.3% |
| `lib/policy.c` | 68.4% → 86.7% |
| `common/wallet.c` | 71.3% → 85.1% |

## 3. Issues found & fixes

| Site | UBSan check | Verdict | Action |
|---|---|---|---|
| `lib/policy.c:1959` `older->n & ~SEQUENCE_LOCKTIME_TYPE_FLAG` | implicit-integer-sign-change (`int -4194305` → `uint32_t 4290772991`) | benign result (mask is correct), real consensus code | **fix** — `~(uint32_t) SEQUENCE_LOCKTIME_TYPE_FLAG` |
| `fuzzing/mock/fuzz_sha256.c` | unsigned overflow + unsigned-shift-base | benign (SHA wraps by design) | **mask** — `__attribute__((no_sanitize("integer")))` |
| `sign_psbt.h:37` `address_index` | implicit-integer-sign-change | benign (BIP32 index, no neg sentinel) | **fix** — field `int` → `uint32_t` (kills 2 sites: `preprocess_outputs.c:78`, `process_in_outs.c:122`) |
| `lib/check_merkle_tree_sorted.c:27` | implicit truncation | benign (bounded size) | **fix** — explicit `(uint32_t)` casts |
| `sign_psbt/preprocess_inputs.c:178,212` | uint64 overflow | **real ordering bug** — `BITCOIN_TOTAL_SUPPLY` guard ran *after* the add | **fix** — validate before each accumulation; dropped redundant trailing check |

Mask = mock only. Fixes = src. Only behavioral change is the amount-overflow ordering
(rejects an impossible over-supply input slightly earlier; valid PSBTs unaffected).

```bash

diff --git a/fuzzing/mock/fuzz_sha256.c b/fuzzing/mock/fuzz_sha256.c
index 58eb0316..4b5375aa 100644
--- a/fuzzing/mock/fuzz_sha256.c
+++ b/fuzzing/mock/fuzz_sha256.c
@@ -34,6 +34,11 @@ static inline void write_be32(uint8_t *p, uint32_t v)
     p[3] = (uint8_t) v;
 }

+// SHA-256 is defined over mod-2^32 arithmetic and 32-bit rotates, so every
+// integer op here wraps / shifts by design. Silence UBSan's opt-in integer
+// checks (unsigned overflow, unsigned-shift-base, truncation) for this
+// function only, so they don't drown out genuine findings elsewhere.
+__attribute__((no_sanitize("integer")))
 static void fuzz_sha256_transform(uint32_t state[8], const uint8_t block[64])
 {
     uint32_t W[64];
```

```bash

diff --git a/src/handler/lib/check_merkle_tree_sorted.c b/src/handler/lib/check_merkle_tree_sorted.c
index 39467045..b09519d4 100644
--- a/src/handler/lib/check_merkle_tree_sorted.c
+++ b/src/handler/lib/check_merkle_tree_sorted.c
@@ -24,10 +24,12 @@ int call_check_merkle_tree_sorted_with_callback(dispatcher_context_t *dispatcher
 
     for (size_t cur_el_idx = 0; cur_el_idx < size; cur_el_idx++) {
         uint8_t cur_el[MAX_CHECK_MERKLE_TREE_SORTED_PREIMAGE_SIZE];
+        // tree size and leaf index are bounded well below 2^32 in practice;
+        // cast to the callee's uint32_t parameters explicitly.
         int cur_el_len = call_get_merkle_leaf_element(dispatcher_context,
                                                       root,
-                                                      size,
-                                                      cur_el_idx,
+                                                      (uint32_t) size,
+                                                      (uint32_t) cur_el_idx,
                                                       cur_el,
                                                       sizeof(cur_el));
 
diff --git a/src/handler/lib/policy.c b/src/handler/lib/policy.c
index 28ea7a7c..89f08f3c 100644
--- a/src/handler/lib/policy.c
+++ b/src/handler/lib/policy.c
@@ -1956,7 +1956,7 @@ static int check_older_node_cb(const policy_node_t *node, void *callback_state)
     (void) callback_state;
     if (node->type == TOKEN_OLDER) {
         const policy_node_with_uint32_t *older = (const policy_node_with_uint32_t *) node;
-        uint32_t n = older->n & ~SEQUENCE_LOCKTIME_TYPE_FLAG;
+        uint32_t n = older->n & ~(uint32_t) SEQUENCE_LOCKTIME_TYPE_FLAG;
         if (n < 1 || n > 65535) {
             return -1;
         }
diff --git a/src/handler/sign_psbt.h b/src/handler/sign_psbt.h
index 6e149aa0..cf19a614 100644
--- a/src/handler/sign_psbt.h
+++ b/src/handler/sign_psbt.h
@@ -34,7 +34,7 @@ typedef struct {
                                 // matched with the current key expression in the signing flow
 
     bool is_change;
-    int address_index;
+    uint32_t address_index;
 
     // For an output, its scriptPubKey
     // for an input, the prevout's scriptPubKey (either from the non-witness-utxo, or from the
diff --git a/src/handler/sign_psbt/preprocess_inputs.c b/src/handler/sign_psbt/preprocess_inputs.c
index 59bfc562..a789f483 100644
--- a/src/handler/sign_psbt/preprocess_inputs.c
+++ b/src/handler/sign_psbt/preprocess_inputs.c
@@ -175,6 +175,12 @@ bool __attribute__((noinline)) preprocess_inputs(
                 return false;
             }
 
+            // sanity check before accumulating, to avoid overflowing the total
+            if (input.prevout_amount > BITCOIN_TOTAL_SUPPLY) {
+                PRINTF("Input amount exceeds Bitcoin total supply!\n");
+                SEND_SW(dc, SW_INCORRECT_DATA);
+                return false;
+            }
             st->inputs_total_amount += input.prevout_amount;
         }
 
@@ -209,6 +215,12 @@ bool __attribute__((noinline)) preprocess_inputs(
                 }
             } else {
                 // we extract the scriptPubKey and prevout amount from the witness utxo
+                // sanity check before accumulating, to avoid overflowing the total
+                if (wit_utxo_prevout_amount > BITCOIN_TOTAL_SUPPLY) {
+                    PRINTF("Input amount exceeds Bitcoin total supply!\n");
+                    SEND_SW(dc, SW_INCORRECT_DATA);
+                    return false;
+                }
                 st->inputs_total_amount += wit_utxo_prevout_amount;
 
                 input.prevout_amount = wit_utxo_prevout_amount;
@@ -217,13 +229,6 @@ bool __attribute__((noinline)) preprocess_inputs(
             }
         }
 
-        if (input.prevout_amount > BITCOIN_TOTAL_SUPPLY) {
-            // sanity check to avoid overflows in amounts
-            PRINTF("Input amount exceed Bitcoin total supply!\n");
-            SEND_SW(dc, SW_INCORRECT_DATA);
-            return false;
-        }
-
         // check if the input is internal; if not, continue
 
         int is_internal = is_in_out_internal(dc, st, sign_psbt_cache, &input.in_out, true);
```

.clusterfuzzlite/Dockerfile
N3ur0sis: Yes, exactly ! Though we could already unfix the ref and only the CFL CI job will fail until sdk is merged.

src/common/merkle.c
N3ur0sis: Indeed, i'll rewrite history cleanly once everything is reviewed and no more changes let to do.

.clusterfuzzlite/Dockerfile
iartemov-ledger: Better to keep the as "at 1st SDK change is merged and then per app onces" I believe ;)

src/common/merkle.c
iartemov-ledger: This commis still needs to be split into:
- https://github.com/LedgerHQ/app-bitcoin-new/pull/502/commits/4cc2e9e30c1311e432a1a14d0a17f0d7853199b4 for `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` part
- the real fix:
```
        // also the number of leaves of the left subtree. Unsigned: depth can reach 32.
        uint32_t mask = 1U << (depth - 1);
```
, does not it ?

N3ur0sis: missed that one indeed :+1:

fuzzing/README.md
```diff
@@ -0,0 +1,279 @@
...
+   array in `mock/wallet_model.c` (ordering is append-only). Pick the
+   right group (see the table above) and fill in the correct `n_keys` /
+   `purposes[]` so `build_key_info` derives distinct xpubs.
+4. **Adjust structured mutation pressure** — tune `btc_field_aware_tweak`
+   in `harness/fuzz_dispatcher.c` and keep the README's slot/fault tables
+   in sync when you add new targeted operations.
+5. **Adjust fuzz-only repairs** — keep repairs gated with
+   `#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` and document any new one in
+   the source site that applies it.
+6. **Promote corpus inputs** — pack a campaign's merged corpus into
```
iartemov-ledger: We may be missing some as "When to update the corpus".

.clusterfuzzlite/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM gcr.io/oss-fuzz-base/base-builder:v1
+
+# Pillow is required by the SDK glyph generator (icon2glyph.py).
+RUN pip3 install --break-system-packages --no-cache-dir pillow>=3.4.0
+
+RUN apt update && apt install -y ninja-build libbsd-dev pkg-config git
+
+ARG LEDGER_SECURE_SDK_REPO=https://github.com/LedgerHQ/ledger-secure-sdk.git
+ARG LEDGER_SECURE_SDK_REF=d8d48fe7fe7446b09a9d3065d460535a07a607dc
```
iartemov-ledger: I believe it should be a moving target - `master` in our case.
Otherwise how will we maintain it:
- it would not very scalable to update all the apps that use the fuzzing ?
- plus what changes in the fuzzing framework of the SDK should trigger the update and who will communicate to app owners?

iartemov-ledger (commented):
The branch has to be rebased.
I was able to rebuild the container and run `"$BOLOS_SDK"/fuzzing/scripts/app-campaign.sh --app-dir /app cov-run`. 
Left a a couple of comments/questions.
Also I'dl like to have a feedback from @bigspider ;)

.clusterfuzzlite/Dockerfile
N3ur0sis: Yes, this was temporary until SDK PR was merged, i've updated it

fuzzing/mock/fuzz_sha256.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+// Lightweight SHA-256 for self-consistent fuzz Merkle commitments (no real crypto).
```
bigspider: Why 'no real crypto'? Isn't this a full implementation of SHA256?

fuzzing/mock/fuzz_varint.h
```diff
@@ -0,0 +1,49 @@
...
+        out[0] = 0xFD;
+        out[1] = (uint8_t) (v & 0xFF);
+        out[2] = (uint8_t) ((v >> 8) & 0xFF);
+        return 3;
+    }
+    out[0] = 0xFE;
+    for (int i = 0; i < 4; i++)
+        out[1 + i] = (uint8_t) ((v >> (8 * i)) & 0xFF);
+    return 5;
+}
```
bigspider: Probably irrelevant for the purposes of this mock, but this is incomplete, as values larger than `0xFFFFFFFF` would be correctly encoded as a 0xFF prefix followed by little-endian 8 bytes.

(it might also be more compact using the fuzzi_write_xxx_le functions below)

bigspider (dismissed):
It needs rebase, otherwise LGTM - I only left a couple of nits.

I reviewed the changes to the app's source code more carefully than the fuzzing framework code, where I skimmed through several parts.

One observation is that there is certainly a substantial amount of duplication between the unit tests and the fuzzing framework related to
- psbt handling
- the semantic_host vs the mock_dispatcher
since they were written concurrently. So there might be a chance to extract common code that is useful for both the unit tests framework and the fuzzing framework. It could be left to a future PR, though.

iartemov-ledger:
FYI `fuzzing` folder added size is ~527KB.

fuzzing/mock/fuzz_varint.h
N3ur0sis: Both points fixed : the 0xFF + little-endian 8-byte branch is there, and it now uses U4LE_ENCODE/U8LE_ENCODE from the SDK's os_utils.h instead of open-coding the bytes.

fuzzing/mock/fuzz_sha256.h
N3ur0sis: You were right that the comment was wrong, it was a real SHA256. It's since been deleted: the harness calls the SDK's cx_hash_sha256 directly (mock/psbt_model.c, mock/wallet_model.c), so there's no second implementation to drift.

fuzzing/base-corpus.zip
iartemov-ledger, line 1: Needs to be updated as getting:
```
error: base corpus /app/fuzzing/base-corpus.zip is incompatible with the current build
```
The update procedure (a note for me):
```
1. enabled = true in fuzz-manifest.toml

2. "$BOLOS_SDK"/fuzzing/scripts/app-campaign.sh --app-dir /app bitcoin-run

3.  python3 "$BOLOS_SDK"/fuzzing/scripts/corpus.py promote \
  /app/.fuzz-artifacts/bitcoin-run/targets/fuzz_app/corpus /app/fuzzing/base-corpus.zip

4. enabled = false in fuzz-manifest.toml and to commit fuzzing/base-corpus.zip
```

?

N3ur0sis: Fixed, the corpus is repromoted. On the procedure, steps 1 and 4 are now gone, i set enable = true by default.  
And something worth knowing : a corpus is tied to the SDK the campaign against. Always good practice to check there is no `base corpus ... is incompatible with the current build`

N3ur0sis:
Rebased on `develop`. @bigspider  your approval was auto-dismissed by the force-push, the PR's changes to src/ are identical to what you approved. The only other difference is unit-tests/libs/mock_dispatcher.c: I handled the semantic_host vs mock_dispatcher duplication you pointed out, so the fuzzing now uses the same mock as the UTs.

bigspider (approved):
LGTM

Issue #354, opened 2025-08-29
Support signing transactions with output amount > input amount


kewde:
PSBTs with custom SIGHASH flags may not always have all the inputs already because we don't know who the counterparty will be yet.
This leads to a situation where the output amount > input amount and the firmware throws an error.

workaround is to add a dummy input - just make sure it's not included in the sighash of the real inputs you wish to sign.
maybe skip the amount check if all inputs have sighashes that support adding more inputs later on? they already get hit with a scary warning about custom sighashes anyways

bigspider:
Addressed in #512

PR #512, opened 2026-06-25
Trustworthy sign_psbt amount/fee display for non-default sighash

### LINKED ISSUE #124: Revisit signing flow and UX in the presence of sighash flags
Currently, if any input is not signed with `SIGHASH_ALL/SIGHASH_DEFAULT`, we show a generic warning.
There are potentially several aspects that might be changed:

- we currently check that `SUM(output_amounts) ≤ SUM(input_amounts)`; that check is too restrictive if there are inputs that don't commit to the entire transaction (e.g.: `SIGHASH_SINGLE`; related issue #123)
- How should the UX reflect and handle those advanced use cases?

One way could be to have an "advanced" transaction validation mode, where all inputs (and their SIGHASH_FLAG if appropriate) and all outputs are shown on-screen.

iartemov-ledger:
sighash_proposal_changes_2

**See the seequences and screens for different sighash combinations :**
- for Flex - 
[sighash_flex_snapshot_sequences.html](https://github.com/user-attachments/files/30123974/sighash_flex_snapshot_sequences.html)

- For Nano S Plus - 
[sighash_flex_snapshot_sequences_nanosp.html](https://github.com/user-attachments/files/30123983/sighash_flex_snapshot_sequences_nanosp.html)

Closes: #124


tests/test_sign_psbt_with_sighash_types.py
```diff
@@ -516,6 +516,36 @@ def test_sighash_single_anyone_output_changed(navigator: Navigator, firmware: Fi
     assert bip0340.schnorr_verify(sighash_bitcoin_core_single_anyone_1, pubkey1, partial_sig1.signature[:-1])
 
 
+def test_sighash_anyonecanpay_negative_fee_shows_receive(navigator: Navigator, firmware: Firmware,
+                                                         client: RaggerClient, test_name: str):
+    # ANYONECANPAY leaves the input set open, so other parties may fund the rest of
+    # the transaction. Here we bump the change output above our own inputs: we end up
+    # net-receiving and inputs < outputs. The device must NOT reject this as a negative
+    # fee, and must show "You receive" with the fee marked unavailable.
+    toggle_nonstandard_sighash_setting(navigator, firmware)
```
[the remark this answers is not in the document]
iartemov-ledger: Fixed!

tests/test_sign_psbt_with_sighash_types.py
```diff
@@ -516,6 +516,36 @@ def test_sighash_single_anyone_output_changed(navigator: Navigator, firmware: Fi
...
+    psbt = open_psbt_from_file(f"{tests_root}/psbt/sighash/sighash-all-anyone-can-pay-sign.psbt")
+    psbt.tx.vout[0].nValue = 9929389  # > inputs total (9919389)
+
+    result = client.sign_psbt(psbt, tr_wallet, None, navigator,
+                              instructions=sign_psbt_instruction_approve(firmware, has_sighashwarning=True),
+                              testname=test_name)
+    assert len(result) == 2
+
```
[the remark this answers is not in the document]
iartemov-ledger: Fixed!

src/ui/display.h
```diff
@@ -8,6 +8,7 @@
 #include "format.h"
```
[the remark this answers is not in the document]
iartemov-ledger: fixed

src/common/sighash.h
```diff
@@ -51,3 +52,33 @@ static inline sighash_class_t classify_sighash(uint32_t sighash_type, int segwit
...
+static inline bool sighash_commits_all_inputs(uint32_t sighash_type) {
+    return (sighash_type & SIGHASH_ANYONECANPAY) == 0;
+}
+
+// True only for base-type ALL (ALL, or DEFAULT on taproot); NONE/SINGLE leave the
+// outputs (and our change) free to change after signing.
+static inline bool sighash_commits_all_outputs(uint32_t sighash_type) {
+    uint32_t base = sighash_type & ~(uint32_t) SIGHASH_ANYONECANPAY;
+    return base == SIGHASH_ALL || base == SIGHASH_DEFAULT;
+}
```
bigspider: I think what we care about in these functions is: _does the signature commit to all the inputs/outputs that are provided in the PSBT that we are signing_? That is because we base the display info based on the provided PSBT. We don't necessarily care if other inputs/outputs are added by someone else, as long as it doesn't invalidate what we committed to.

Therefore, we might generalize these functions a bit:
- if there is only 1 input, then `sighash_commits_all_inputs` could always return true. All the sighash flags  commit to the current input, whether or not ANYONECANPAY is present.
- if there is only 1 input and output, then SIGHASH_SINGLE still commits to that all outputs, as trivially there is only one.

In particular, this would allow covering the SINGLE|ACP case among the 'safer' patterns (the one-input-one-output case is interesting as it is already used in some protocols). Note that we still don't know the fee in this case (and someone else might be paying for it!), but we do know exactly how much we're spending/receiving.

bigspider (changes requested):
Great initiative, I think this is going in a good direction.

As discussed offline, there are a few things that might deserve further thinking:
- It seems hopeless to try to provide a meaningful display when different inputs have different signing flags; but it is unclear if there are use cases for that. We could either forbid or automatically consider unsafe any such scenario (TBD), and focus on cases where all inputs have the same sighash flag (_TL;DR: no sighash flag mixing_)
- SIGHASH_SINGLE actually seems safe in the 1-input-1-output case, and since there are already deployed usecases for it, it might be worth including it
- The UX in the `TX_DISPLAY_SPENT_ONLY` case is not very clear, as it doesn't clearly distinguish between: (a) the amount spent/received from/to the wallet policy; (b) the amount going to each output. This needs to be clearly for the user.
- When not signing with SIGHASH_ALL, but with no sighash flag mixing, it might be worth clearly showing _what's the sighash flag_ used.

For anything not fitting these patterns, it is probably fine to fall back to `TX_DISPLAY_UNAVAILABLE` and just admit that we can't clear-sign it. Maybe the warning could be worded to be more explicitly scarier.

Possible add-ons that are fine to defer to future PRs:
- Even in the `TX_DISPLAY_UNAVAILABLE` case, if a certain registered wallet policy is involved in a transaction
- An advanced transaction mode showing info also about each input (e.g.: if internal, its address, its amount, the sighash flag requested for signing if non-default).

I think only small changes are needed to fill these gaps, and it will make the display for such transactions a lot more meaningful.

src/common/sighash.h
iartemov-ledger: Generalizing `sighash_commits_all_inputs()` to return true for a single input would be unsafe, because this function isn't just informational - it feeds `sighash_inputs_open`, which drives the negative-fee check skip in `preprocess_outputs()`.

The subtlety is that two different ideas are being mixed up here:

- "commits to the inputs provided in the PSBT" -  yes,  true for 1 input
- "the input set is closed, i.e. no one can add inputs" = `!ANYONECANPAY` - what this function actually means

A single `ANYONECANPAY `input satisfies the first but not the second: others can still append inputs. If `sighash_commits_all_inputs()` returned true there, `sighash_inputs_open` would become false, the negative-fee check would no longer be skipped, and a valid 1-input `ANYONECANPAY` maker PSBT (output > our input) would be wrongly rejected.

bigspider: Sure, the negative-fee check should be only done when the fee can be computed, which is basically only the case for SIGHASH_ALL/DEFAULT. Other than that, I don't think there is any other problem?

iartemov-ledger:
Normally the change-requests expressed all along the PR work have been addressed or postponned:

# sign_psbt non-default-sighash — change requests & resolution

Branch: `negative_fee` (vs `develop`). "Done" = implemented in this PR; "Postponed" = out of scope, tracked for a follow-up.

| Change request | How it's addressed in the PR |
|---|---|
| Trustworthy amount/fee display for non-default sighash | **Done** — 3 review modes: `FULL` / `NET_ONLY` / `UNAVAILABLE`, per what the signature really commits to |
| **[#512]** Mixed sighash across signed inputs can't be shown coherently | **Done** — mixed sighash → `UNAVAILABLE` |
| **[#512]** A non-default sighash must never show a trusted fee | **Done** — never reaches `FULL`; shows the net "You spend/receive" + "Fees: Not available" instead |
| **[#512]** `SINGLE` / `SINGLE\|ANYONECANPAY` at 1-input / 1-output (coinjoin **maker/taker**) | **Done** — shown as `NET_ONLY`, not `UNAVAILABLE` |
| **[#354]** Sign when output amount > input amount (open input set), incl. **PayJoin** net-receive | **Done** — accepted when the set is open (ACP/NONE/SINGLE), no dummy input; net-receive shown as "You receive". Default sighash still rejects a genuine negative fee |
| **[#520 ]** **PayJoin** display for default `SIGHASH_ALL` — "You receive" + total value of external inputs | **Postponed** — follow-up PR |
| Show the effective sighash on the review | **Done** — "Signing rule" row (`ACP \| ALL`, `SINGLE`, `Mixed`, …) |
| Show the account (with dynamic direction) in a trust-reduced review | **Done** — account row in `NET_ONLY` and `UNAVAILABLE`, from our internal inputs: default → derived label (e.g. `Native SegWit #0`), registered → its name; `From` when spending, `To` when receiving, neutral when unknown |

src/handler/sign_psbt/preprocess_inputs.c
```diff
@@ -105,6 +105,18 @@ void input_keys_callback(dispatcher_context_t *dc,
     }
 }
 
+// Track the sighash across signed inputs (DEFAULT canonicalized to ALL); disagreement
+// -> sighash_mixed.
+static void record_signed_sighash(sign_psbt_state_t *st, uint32_t sighash_type) {
```
bigspider: The word 'signed' made me think that this would be used in the signing part of sign_psbt, not in `preprocess_inputs`.

Maybe something like `track_seen_sighash`?
Flag names could similarly be `seen_` instead of `signed_`.

src/common/sighash.h
```diff
@@ -51,3 +52,41 @@ static inline sighash_class_t classify_sighash(uint32_t sighash_type, int segwit
...
+
+// What the review can show, given the signed inputs' commitment and amount trust.
+typedef enum {
+    TX_DISPLAY_FULL,         // outputs + fee (and total_spent)
+    TX_DISPLAY_NET_ONLY,     // total_spent only; fee unknown
+    TX_DISPLAY_UNAVAILABLE,  // nothing about amounts/fee is reliable
+} tx_display_mode_t;
+
+// FULL if the fee is trusted; NET_ONLY if only the net is exact; else UNAVAILABLE.
```
bigspider: Given that this is a delicate topic, it might be worth documenting the exact meaning of the parameters.

src/handler/sign_psbt/transaction_display.c
```diff
@@ -204,13 +205,90 @@ static bool __attribute__((noinline)) display_warnings(dispatcher_context_t *dc,
...
+    tx_display_mode_t mode = TX_DISPLAY_FULL;
+    if (st->warnings.non_default_sighash) {
+        if (st->sighash_mixed) {
+            mode = TX_DISPLAY_UNAVAILABLE;  // signed inputs disagree: nothing coherent
+        } else {
+            // uniform non-default sighash never fixes the whole set => fee never trustworthy
+            bool fee_trustworthy = false;
+            bool outputs_committed =
+                sighash_commits_provided_outputs(st->signed_sighash, st->n_outputs);
+            bool amounts_trustworthy = !st->warnings.missing_nonwitnessutxo;
```
bigspider: Considering the amounts not-trustworthy if the nonwitnessutxo is missing is a big change and should not be part of this PR.

We already fail for legacy if the nonwitnessutxo is missing, and it is not necessary on taproot.

For segwit, we already give a warning, and that's enough. The attack is remote and unlikely, so we should stick to the FULL display. This situation actually occurs in practice to people (e.g. recovering a wallet from a pruned node, so their node doesn't have the full transaction history and can't fill the nonwitnessutxo) - in fact, [release 2.0.6](https://github.com/LedgerHQ/app-bitcoin/blob/develop/CHANGELOG.md#206---06-06-2022) was done after people complained for this!

So I think `decide_tx_display_mode` should not have the `amounts_trustworthy` argument, and only decide based on the sighash flags.

src/handler/sign_psbt/transaction_display.c
```diff
@@ -220,28 +298,53 @@ bool __attribute__((noinline)) display_transaction(
...
+        bool show_self_transfer_row = is_self_transfer && mode != TX_DISPLAY_NET_ONLY;
+
         /** TRANSACTION CONFIRMATION */
         /* Init*/
         ui_transaction_simplified_init(
-            st->account.is_default ? NULL : st->account.wallet_header.name,
-            is_self_transfer ? 1 : st->n_external_outputs,
-            st->warnings);
+            account_review_label(st, mode, account_label_buf, sizeof(account_label_buf)),
+            is_self_transfer ? (show_self_transfer_row ? 1 : 0) : st->n_external_outputs,
```
bigspider: not a big fan of nesting `?:`; perhaps worth computing the argument explicitly before the call?

src/handler/sign_psbt/transaction_display.c
```diff
@@ -277,8 +379,11 @@ bool __attribute__((noinline)) display_transaction(
```
bigspider, line 384: Quite an edge case, but the `else` case (more than `MAX_EXT_OUTPUT_SIMPLIFIED_NUMBER` outputs) does not reflect the changes that the main branch has. It would be nice if the UX degrades, but it always remains functionally equivalent.

If not, we should put some thoughts into what's a reasonable behavior, and document it properly. It might be OK to only support ALL/DEFAULT in this branch and keep it simple, tbh.

src/ui/display_nbgl.c
```diff
@@ -54,12 +54,97 @@ const char GA_LOADING_TRANSACTION[] = "Loading transaction";
...
 
-#define N_UX_PAIRS 51
+// Non-default-sighash transaction summary labels (trustworthy-or-bust display)
+const char GA_FEE_NOT_AVAILABLE[] = "Not available";
+const char GA_YOU_SPEND[] = "You spend";
+const char GA_YOU_RECEIVE[] = "You receive";
+const char GA_AMOUNTS_UNAVAILABLE_TITLE[] = "Amounts & fees";
+#ifdef SCREEN_SIZE_WALLET
+const char GA_AMOUNTS_UNAVAILABLE[] =
+    "Cannot be verified for these signing rules. Only sign if you trust the source.";
```
bigspider: "Only sign if you trust the source" is a bit too weak as a warning. You're trusting the software wallet as well, and the hardware can't really protect you from malware tampering with the transaction. So it should be considered as dangerous as blind signing.

Proposal, if it fits on screen:

```suggestion
    "Cannot be verified for these signing rules. Only sign if you expected it and fully trust the software wallet.";
```

src/ui/display_nbgl.c
```diff
@@ -54,12 +54,97 @@ const char GA_LOADING_TRANSACTION[] = "Loading transaction";
...
+#else
+const char GA_AMOUNTS_UNAVAILABLE[] = "Cannot be verified\nReject if not sure";
```
bigspider, line 67: Perhaps we can use here the same long message even on Nano S+/X? The user would have to scroll multiple screens, but I think that's fine for such a dangerous (and anyway unlikely) action.

src/ui/display_nbgl.c
```diff
@@ -54,12 +54,97 @@ const char GA_LOADING_TRANSACTION[] = "Loading transaction";
...
+// Account row label: direction (From/To/neutral) + type. "default" = a standard derivation,
+// "registered" = a registered policy. The type also keeps this "To" distinct from output "To".
+static const char *account_role_label(account_role_t role, bool is_default) {
+    switch (role) {
+        case ACCOUNT_ROLE_TO:
+            return is_default ? "To default" : "To registered";
+        case ACCOUNT_ROLE_NEUTRAL:
+            return is_default ? "Default" : "Registered";
+        default:
+            return is_default ? "From default" : "From registered";
```
bigspider: Seeing the screenshots, I think default/registered is not very clear (_what_ is registered or default?).

We can try adding the word 'account' explicitly:

```
    switch (role) {
        case ACCOUNT_ROLE_TO:
            return is_default ? "To default account" : "To registered account";
        case ACCOUNT_ROLE_NEUTRAL:
            return is_default ? "Default account" : "Registered account";
        default:
            return is_default ? "From default account" : "From registered account";
```

It doesn't  not fit on small screens, though.

Maybe on small screens we just say "To account"/"Account"/"From account", which loses the clean separation between registered or default ones, but it is still much easier to understand.

src/ui/display_nbgl.c
```diff
@@ -54,12 +54,97 @@ const char GA_LOADING_TRANSACTION[] = "Loading transaction";
...
+        case SIGHASH_NONE:
+            return GA_SIGHASH_NONE;
+        case SIGHASH_SINGLE:
+            return GA_SIGHASH_SINGLE;
+        case SIGHASH_ANYONECANPAY | SIGHASH_ALL:
+            return GA_SIGHASH_ACP_ALL;
+        case SIGHASH_ANYONECANPAY | SIGHASH_NONE:
+            return GA_SIGHASH_ACP_NONE;
+        case SIGHASH_ANYONECANPAY | SIGHASH_SINGLE:
+            return GA_SIGHASH_ACP_SINGLE;
```
bigspider: Not sure global defines is cleaner than just inlining all the strings here, in this case?
The constants aren't used elsewhere, and are unlikely to be.

src/handler/sign_psbt/transaction_display.c
bigspider: Not caused by this PR, but the comment is obsolete (should say "more than `MAX_EXT_OUTPUT_SIMPLIFIED_NUMBER` outputs").

bigspider (changes requested):
Good progress!
I commented on a number of issues here and there, most of them small - the only major one being the overly tight treatment of `amounts_trustworthy` which IIUC would be a regression that impacts users.

src/handler/sign_psbt/transaction_display.c
iartemov-ledger: Normally the streaming branch (more than `MAX_EXT_OUTPUT_SIMPLIFIED_NUMBER` external outputs) has since been aligned with the simplified one, so it stays functionally equivalent:

- **Context page** (`ui_transaction_streaming_prompt`): account row + "Signing rule", same as the simplified flow.
- **Per-output pages**: address + amount, as before.
- **Final page** (`ui_display_transaction_streaming_flow`):
  - `FULL` → `Fees: <fee>` (+ high-fee warning), identical to main.
  - `NET_ONLY` → "You spend/receive" + "Fees: Not available".

What's actually reachable here:

- `FULL` (ALL/DEFAULT): full display, same as main.
- `NET_ONLY`: only via `ACP|ALL` (base ALL keeps the outputs committed, ANYONECANPAY opens the fee). SINGLE-with-1-output can't occur with many outputs.
- `UNAVAILABLE` (NONE / SINGLE-multi / mixed): **never reaches this branch** — `display_transaction` routes it to the simplified notice flow before the output-count split, regardless of the number of outputs.

So it always signs, always warns, and shows the strongest coherent amount info the sighash allows.

src/ui/display_nbgl.c
iartemov-ledger: Without deeper changes it does not create several screens to scroll, but renders it as follows:

src/handler/sign_psbt/preprocess_inputs.c
iartemov-ledger: Done!

src/common/sighash.h
iartemov-ledger: Done!

src/handler/sign_psbt/transaction_display.c
iartemov-ledger: Done!

src/ui/display_nbgl.c
iartemov-ledger: Done!

src/ui/display_nbgl.c
iartemov-ledger: Done!

src/ui/display_nbgl.c
iartemov-ledger: Done!

src/handler/sign_psbt/transaction_display.c
iartemov-ledger: Done!

src/handler/sign_psbt/transaction_display.c
iartemov-ledger: Done!

src/ui/display_nbgl.c
bigspider: Ok, I guess it's less informative but 'cannot be verified' should be scary enough.

src/handler/sign_psbt/transaction_display.c
bigspider: Thanks - on a second read, that indeed seems to be the case.

src/ui/display.h
```diff
@@ -113,7 +131,14 @@ typedef struct {
     char address_or_description[MAX_EXT_OUTPUT_SIMPLIFIED_NUMBER]
                                [MAX(MAX_ADDRESS_LENGTH_STR + 1, MAX_OPRETURN_OUTPUT_DESC_SIZE)];
     char amount[MAX_EXT_OUTPUT_SIMPLIFIED_NUMBER][MAX_AMOUNT_LENGTH + 1];
-    char fee[MAX_AMOUNT_LENGTH + 1];
+    char fee[MAX_AMOUNT_LENGTH + 1];  // formatted: network fee, or |total_spent| for NET_ONLY
```
bigspider: This field is now used for two different things. If kept like this, then it should be renamed to something like `fee_or_netamount`, but maybe it's even better to have two separate fields? Post Nano-S, no need to hyper-optimize memory usage...

src/handler/sign_psbt/transaction_display.c
```diff
@@ -220,28 +297,59 @@ bool __attribute__((noinline)) display_transaction(
...
     if (st->n_external_outputs <= MAX_EXT_OUTPUT_SIMPLIFIED_NUMBER) {
         // A simplified flow for most transactions: show it using the classical review if there is
         // exactly 0 (self-transfer) or <= MAX_EXT_OUTPUT_SIMPLIFIED_NUMBER external outputs to show
         // to the user
 
         bool is_self_transfer = st->n_external_outputs == 0;
 
+        // In NET_ONLY "You spend/receive" already states the amount; a "0 (self-transfer)"
+        // row would contradict a net-receive, so omit it.
+        bool show_self_transfer_row = is_self_transfer && mode != TX_DISPLAY_NET_ONLY;
```
bigspider: I think it's more precise to write:

```suggestion
        // With DISPLAY_FULL and no external outputs, we add a line to make it clear that
        // the amount sent is 0 (except the fee that is shown separately).
        bool show_self_transfer_row = is_self_transfer && mode == TX_DISPLAY_FULL;
```

since the self-transfer line also wouldn't make any sense for the screens that are currently under TX_DISPLAY_UNAVAILABLE, even if that were to change in the future.

src/ui/display_nbgl.c
```diff
@@ -175,12 +263,15 @@ void ui_display_transaction_simplified_flow_add(void) {
...
+        bool force_output_page =
+            state->display_mode != TX_DISPLAY_FULL && state->n_outputs == 1 && output_index == 0;
         pairs[n_pairs++] = (nbgl_layoutTagValue_t) {.item = "Amount",
                                                     .value = state->amount[output_index],
-                                                    .forcePageStart = false};
+                                                    .forcePageStart = force_output_page};
 
         pairs[n_pairs++] = (nbgl_layoutTagValue_t) {
-            .item = "To",
+            .item = "Address",
```
bigspider: This changes the display for external outputs in normal transactions from "To" to "Address", which we don't want.

src/ui/display.h
```diff
@@ -102,6 +104,22 @@ typedef struct {
     char signer_index[sizeof("Key @999 <theirs>")];
 } ui_cosigner_pubkey_and_index_state_t;
 
+// Which side of the transaction the wallet account is on, for the account review row.
+typedef enum {
+    ACCOUNT_ROLE_FROM,     // net spend (and the default/FULL case)
+    ACCOUNT_ROLE_TO,       // net receive
+    ACCOUNT_ROLE_NEUTRAL,  // direction not knowable (UNAVAILABLE)
```
bigspider: Nit: neutral might seem to imply that the the net amount is 0l `ACCOUNT_ROLE_UNKNOWN` or `ACCOUNT_ROLE_UNKNOWABLE` would be more precise.

bigspider (changes requested):
Mostly nits at this point, except the change "To" ==> "Address" which shouldn't happen for normal transactions.

src/ui/display_nbgl.c
```diff
@@ -54,12 +54,104 @@ const char GA_LOADING_TRANSACTION[] = "Loading transaction";
...
+#ifdef SCREEN_SIZE_WALLET
+const char GA_AMOUNTS_UNAVAILABLE[] =
+    "Cannot be verified for these signing rules. Only sign if you expected it and fully trust the "
+    "software wallet.";
+#else
+const char GA_AMOUNTS_UNAVAILABLE[] = "Cannot be verified\nReject if not sure";
+#endif
+const char GA_SIGNING_RULE_TITLE[] = "Signing rule";
+
+// +1 slot for the "Signing rule" row shown in the non-default-sighash flows
```
bigspider: This comment doesn't quite make sense in the code, as it explains why it was increased from 51 to 52.

src/ui/display.h
iartemov-ledger: Done.

src/handler/sign_psbt/transaction_display.c
iartemov-ledger: Done.

src/ui/display.h
iartemov-ledger: Done.

src/ui/display_nbgl.c
iartemov-ledger: Done + few other places.

src/ui/display_nbgl.c
iartemov-ledger: Reverted in this PR.
Streamed review ill be aligne to "Address->To" in a separate PR.

bigspider (approved):
LGTM

PR #514, opened 2026-07-02
app-bitcoin-new rename to app-bitcoin - let's now use the docker image - already deployed after the push on develop

iartemov-ledger:


iartemov-ledger:
Merging without approval as a logical follow-up of https://github.com/LedgerHQ/app-bitcoin/pull/513 to restore normal BTC GH/CI functionality.

PR #528, opened 2026-07-17
Display external input amounts and net-spend

### LINKED ISSUE #520: Acknowledge in UI amount of external inputs when present
## What's missing after https://github.com/LedgerHQ/app-bitcoin/pull/512
The trust-reduced display (`You spend` / `You receive`, `NET_ONLY`) triggers **only for
non-default sighash** (`ANYONECANPAY`, open input set). For a **default `SIGHASH_ALL`**
transaction the review is `FULL` (outputs + real fee) - correct and trustworthy, but it does
**not** surface the PayJoin angle even when someone else co-funds the transaction.

## The PayJoin / SIGHASH_ALL case
A coordinator can build a **fully-committed** (`SIGHASH_ALL`) transaction that already contains
**external inputs** (the counterparty's) alongside yours. From the signer's side:
- the tx is complete and the fee is known (→ still `FULL`), but
- part of the funding comes from someone else, and your **net** balance change may even be a
  **receive** (receive amount > spent amount).

## Proposed enhancement (separate PR)
In `FULL` mode, when external inputs are present:
1. Show **"You receive"** (net) when the net effect is positive - i.e. `total_spent < 0`,
   same criterion already used in `NET_ONLY`.
2. Show the **total value of external inputs** (the counterparty's contribution), so the user
   sees how much of the transaction someone else funded.

## Why it's cheap to add later
The building blocks already exist:
- `total_spent = internal_inputs_total_amount − change_total_amount` (sign gives spend/receive);
- external inputs total = `inputs_total_amount − internal_inputs_total_amount`
  (`inputs_total_amount` already counts every input; `n_external_inputs` is tracked);
- `warnings.external_inputs` already flags this situation.

So it's mainly a `FULL`-mode display extension, not new parsing.

## Why parked
`FULL` (default sighash) is already accurate and safe; this is a **UX enhancement** for the
PayJoin experience, not a correctness fix - so it belongs in its own PR rather than the
non-default-sighash trust-reduction change.

bigspider:
Prior to this PR, the app warns of the presence of external inputs, but otherwise ignores them. While the amount shown for the (external) outputs is correct, this doesn't give a complete picture of the transaction being signed:
- no info about how much the external inputs are contributing
- no info about how much is the 'flow of funds' in this transaction with respect to the wallet policy involved in this PSBT.

Example:
<p>
  Screenshot from 2026-07-20 14-52-09
  Screenshot from 2026-07-20 14-52-13
  Screenshot from 2026-07-20 14-52-16
  Screenshot from 2026-07-20 14-52-20
</p>

After the fix, two additional fields are shown for such transactions: the total amount of external inputs, and "You spend"/"You receive", similar to what #512 did for transactions with sigash flags: 

<p>
  Screenshot from 2026-07-20 14-49-44
  Screenshot from 2026-07-20 14-49-52
  Screenshot from 2026-07-20 14-49-55
  Screenshot from 2026-07-20 14-49-56
  Screenshot from 2026-07-20 14-49-58
</p>

The 'External inputs amount' pair is also shown in the presence of sighash flags, as long as the input set is committed (that is, not `ANYONECANPAY`).

Nothing changes for transactions with SIGHASH_ALL/DEFAULT and no external inputs.

On NanoS+/NanoX, the string `External inputs amount` doesn't fit in a line, so the shorter `External amounts` is used:

<p>
Screenshot from 2026-07-23 11-01-19
Screenshot from 2026-07-23 11-01-25
Screenshot from 2026-07-23 11-01-26
Screenshot from 2026-07-23 11-01-34
Screenshot from 2026-07-23 11-01-36
Screenshot from 2026-07-23 11-01-38
Screenshot from 2026-07-23 11-01-40
Screenshot from 2026-07-23 11-01-42
Screenshot from 2026-07-23 11-01-47
Screenshot from 2026-07-23 11-01-49
Screenshot from 2026-07-23 11-01-51
</p>

Closes: #520


src/ui/display_nbgl.c
```diff
@@ -153,6 +161,28 @@ static unsigned int append_net_only_pairs(const ui_validate_transaction_state_t
...
+                                            .forcePageStart = force_page};
+    return idx;
+}
+
+// Appends, for a FULL-mode transaction with external inputs, the "External inputs amount" row
+// and the net "You spend/receive" row. The trustworthy "Fees" row is still added by the caller.
+static unsigned int append_external_inputs_pairs(const ui_validate_transaction_state_t *state,
+                                                 unsigned int idx,
+                                                 bool force_page) {
+    idx = append_external_amounts_row(state, idx, force_page);
```
iartemov-ledger, line 179: Claude suggests the following additional checking for the situation when for some reason in future we reclassify `ANYONECANPAY` as "safe" and it reaches `FULL` mode (as in `FULL` mode the code gates the "External inputs amount" row on `has_external_inputs`, yet `state->unverified_inputs` is only formatted when `show_external_inputs_amount` is true ([display.c:301] - so it can become a stale buffer):
```
diff --git a/src/ui/display_nbgl.c b/src/ui/display_nbgl.c
index 18a57753..302c1c82 100644
--- a/src/ui/display_nbgl.c
+++ b/src/ui/display_nbgl.c
@@ -176,10 +176,16 @@ static unsigned int append_external_amounts_row(const ui_validate_transaction_st
 static unsigned int append_external_inputs_pairs(const ui_validate_transaction_state_t *state,
                                                  unsigned int idx,
                                                  bool force_page) {
-    idx = append_external_amounts_row(state, idx, force_page);
+    // Only emit the external-inputs total when it was formatted (show_external_inputs_amount);
+    // otherwise the buffer is stale. The net "You spend/receive" row is always safe to show.
+    if (state->show_external_inputs_amount) {
+        idx = append_external_amounts_row(state, idx, force_page);
+        force_page = false;  // the external row already started the page
+    }
     pairs[idx++] =
         (nbgl_layoutTagValue_t) {.item = state->spent_is_receive ? GA_YOU_RECEIVE : GA_YOU_SPEND,
-                                 .value = state->net_amount};
+                                 .value = state->net_amount,
+                                 .forcePageStart = force_page};
     return idx;
 }
```

src/handler/sign_psbt/transaction_display.c
```diff
@@ -261,6 +261,19 @@ bool __attribute__((noinline)) display_transaction(
     int64_t total_spent =
         (int64_t) st->internal_inputs_total_amount - (int64_t) st->outputs.change_total_amount;
 
+    // Total of the external (unverified) inputs. When present, the outputs and fee alone don't tell
+    // the user how much really leaves (or enters) their wallet, so we additionally show this amount
+    // and the net `total_spent`. (inputs_total_amount always includes
+    // internal_inputs_total_amount.)
+    bool has_external_inputs = st->warnings.external_inputs;
+    uint64_t external_inputs_amount = st->inputs_total_amount - st->internal_inputs_total_amount;
```
iartemov-ledger, line 270: Got following warning from  `SECURITY.REVIEW.md` pass - what do you think?

### ⚠️ WARNING — "External inputs amount" / fee rest on uncommitted host data for segwit-v0 internal inputs

**Severity:** Warning (accuracy / defense-in-depth) — **pre-existing**, not introduced by this branch
**File:** `src/handler/sign_psbt/transaction_display.c:269-275`
**CWE:** CWE-345 (Insufficient Verification of Data Authenticity)

**Description**

The "External inputs amount" total and the FULL-mode fee both include external-input
amounts taken from host-supplied witness-UTXO values. For **external (unsigned) inputs**,
the device never verifies these amounts against the real prevouts.

When the *internal* (signed) inputs are **segwit-v0** (BIP143 commits only each input's
*own* amount), a malicious host can report false external amounts while the internal
signatures stay valid — so the displayed external total / fee can be inaccurate **even
with a closed input set**. For **taproot** internal inputs, BIP341 `sha_amounts` commits
*all* input amounts, so the total is genuinely trustworthy.

The `show_external_inputs_amount` gate keys off input-set openness (`!sighash_inputs_open`,
i.e. no ANYONECANPAY), **not** amount-commitment. It therefore slightly over-claims
trustworthiness in the segwit-v0-internal case.

```c
// transaction_display.c:275
bool show_external_inputs_amount = has_external_inputs && !st->sighash_inputs_open;
```

**Impact — limited**

- The net **"You spend/receive"** figure is computed only from verified internal inputs and
  verified change (`internal_inputs_total_amount − change_total_amount`), independent of any
  host-supplied external amount — so the user's **actual fund movement is always displayed
  correctly**. No fund-loss / clear-signing bypass.
- The external total and fee share the exact trust level of the fee already shown before this
  branch, so no *new* trust gap is introduced; adding the trustworthy net figure is a net
  transparency improvement.

**Suggested action** (not merge-blocking)

- (a) Minimal: add a comment at the gate noting the total is only amount-committed for taproot
  internal inputs; **or**
- (b) Strict: tighten the gate to also require the amounts-committed condition (e.g. all signed
  inputs are taproot with a non-ANYONECANPAY sighash) before treating the external total as
  trustworthy.

iartemov-ledger (commented):
Good! Just 2 Claude inspired comments.

src/ui/display_nbgl.c
bigspider: Not sure this makes sense, as ANYONECANPAY inherently means that the we're not committing to the other inputs. So it can never be FULL.

The function name is `append_external_inputs_pairs`, so it seems rather weird to check if the external inputs should be shown!

src/handler/sign_psbt/transaction_display.c
bigspider: This was actually a good point, thanks!

Fixed in 5174611d3724cc42bc951dc432d104ca9b543e6f, and snapshots updated in c50a49eb70573853a15bf84b506f7de541ab4a06.

src/handler/sign_psbt/transaction_display.c
```diff
@@ -268,11 +269,20 @@ bool __attribute__((noinline)) display_transaction(
...
-    // NET_ONLY too.
-    bool show_external_inputs_amount = has_external_inputs && !st->sighash_inputs_open;
+    // The external inputs total (and the fee, consequently) can only be trusted when those amounts
+    // are committed to in the signatures: the input set must be closed (no ANYONECANPAY, else
+    // more inputs can be appended after signing) and either we sign taproot (BIP341 commits
+    // sha_amounts over every input) or each external amount was hash-verified via its
+    // non-witness-utxo.
+    // Otherwise a malicious host could forge them via an unverified witness-utxo while the
+    // transaction stays valid (the external party signs with the real one).
+    bool signing_taproot = (get_policy_segwit_version(st->account.policy_map) == 1);
```
iartemov-ledger, line 279: Yet 2 comments related to future `baseapp` rebase:

1. This call will crash on `basapp` branch as `get_policy_segwit_version()` dereferences `policy->type` and on `baseapp` branch  `has_no_wallet_policy ⇒ policy_map == NULL ⇒ crash`.
Would need to add a conditional call. 


2. Complex suggestion from Claude:

> Policy-taproot ≠ taproot-signature. `signing_taproot` is derived from the policy type, not from whether the device actually produces a non-ANYONECANPAY taproot signature. On the standard app that's safe (≥1 internal taproot input ⇒ real taproot sig commits `sha_amounts`; 0 internal inputs ⇒ 0 signatures ⇒ a forged external amount buys the attacker nothing). On `baseapp`, a derived app can sign custom, possibly non-taproot inputs while the base policy is `tr(...)` with zero base-internal inputs — then `signing_taproot` is true but no `sha_amounts`-committing signature is produced, re-opening the exact W1 gap. For the port, the check should be "we actually emit ≥1 non-ACP taproot signature over an input in this tx," not "the policy is taproot."

src/handler/sign_psbt/transaction_display.c
iartemov-ledger: Fix is correct and tested.

src/ui/display_nbgl.c
iartemov-ledger: OK, if "ANYONECANPAY inherently means that the we're not committing to the other inputs. So it can never be FULL."  holds forever then fine!

iartemov-ledger (commented):
Yet 2 comments related to future `baseapp` rebase.

iartemov-ledger (approved):
As discussed offline https://github.com/LedgerHQ/app-bitcoin/pull/528#discussion_r3646366866 can be fixed in the `baseapp` branch directly.

PR #532, opened 2026-07-20
Add documentations for sighash and external output features; prepare release 2.5.0

bigspider:
On top of #528.

Add docs for the new features, and bumps the version to 2.5.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,16 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 Dates are in `dd-mm-yyyy` format.
 
-## [2.X.X] - XX-XX-2026
-
-### Added
-
-- Enable stack protector
+## [2.5.0] - 24-07-2026
 
 ### Changed
 
 - Improved the registration UX for multisig and many taproot miniscript wallet policies by showing the *clear-text, human-readable* meaning of the policy. For simple multisig policies, the clear-text description replaces the descriptor template. For a large class of supported taproot miniscript policies, it is shown *in addition to*, and *before*, the raw descriptor template. This lets the user understand the policy while still being able to compare the template against their backup.
+- Signing transactions using non-standard sighash flags now requires enabling the corresponding persistent application setting.
+- Improved the signing display for transactions using non-default sighash flags: the effective signing rule is shown, and the trusted screen now only presents amounts that the signatures actually commit to, falling back to the net amount spent/received (or a warning when nothing can be shown reliably) instead of a potentially misleading breakdown.
+- Improved the signing display for transactions with external inputs: in addition to the existing warning and external outputs, the trusted screen now shows the total amount of the external inputs when it can be reliably determined, and the net amount spent from or received into the account.
 - Standard BIP-87 xpubs can now be exported without explicit user confirmation.
+- Enabled stack protector.
 
 ## [2.4.6] - 30-03-2026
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,16 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-
-### Added
-
-- Enable stack protector
+## [2.5.0] - 24-07-2026
 
 ### Changed
 
 - Improved the registration UX for multisig and many taproot miniscript wallet policies by showing the *clear-text, human-readable* meaning of the policy. For simple multisig policies, the clear-text description replaces the descriptor template. For a large class of supported taproot miniscript policies, it is shown *in addition to*, and *before*, the raw descriptor template. This lets the user understand the policy while still being able to compare the template against their backup.
+- Signing transactions using non-standard sighash flags now requires enabling the corresponding persistent application setting.
```
iartemov-ledger, line 15: Do "non-standard" and "non-default" mean actually the same thing in app-bitcoin context?

iartemov-ledger, line 15: Some of the element here after can probably be put under ###Added category rather than ###Changed?

doc/features.md
```diff
@@ -73,6 +73,39 @@ Taproot policies may use `musig()` key expressions to aggregate several keys int
...
+
+The application *cannot* prove with certainty that such inputs are actually external, since
+malicious software wallets might (for example) not reveal the necessary BIP-32 derivations
+in the PSBT.
+
+While the application only signs for the inputs that are proven as internal to the wallet policy
+involved in the spend, certain attacks remain possible where a user is tricked into spending more
+than expected by approving multiple signing requests. Therefore, the application shows a
+**warning** when attempting to sign such transactions. Users **must not approve** such
+transactions unless they fully understand the warning and its consequences, and were expecting it.
```
iartemov-ledger: Maybe better:

> Users **must not approve** such a transaction unless they fully understand the warning and its consequences, and were already expecting it.

?

doc/integration.md
```diff
@@ -132,6 +132,67 @@ signature may be returned for that input.
...
+  inherently dangerous.
+
+For any non-default sighash, a **Signing rule** row additionally names the effective flag (for
+example `SINGLE`, `ACP | ALL`, or `Mixed`).
+
+### External inputs
+
+An *external input* is an input that the app cannot verify as belonging to the signing wallet
+policy. The app always **warns** about external inputs and, so the user can gauge the transaction's
+real effect, additionally shows:
```
iartemov-ledger: Maybe following is bit more comprehensible:

> The app always **warns** about external inputs and additionally shows the following, so the user can gauge the transaction's real effect:

?

iartemov-ledger (commented):
Few minor comments.

CHANGELOG.md
bigspider: More or less: ALL=1 and DEFAULT=0 have the same effect (but only ALL should be used pre-taproot).
So I preferred to use non-standard here because it means "anything that is not ALL or DEFAULT".

bigspider: As much as I like the cleartext UX, it doesn't let the user do anything new, so it feels more like a 'change to existing features'.

doc/features.md
bigspider: Included in e990d95c22394eed9c4aa9184913bc31cfe4782b.

doc/integration.md
bigspider: Done in e990d95c22394eed9c4aa9184913bc31cfe4782b.

I also added the new explanations about the precise behavior with external outputs, since it changed upstream.
(Will rebase once #528 is merged)

iartemov-ledger (dismissed):
Looks good!

iartemov-ledger (approved):

PR #533, opened 2026-07-21
Harden and refactor PSBT accessor API

bigspider:
This makes several improvements:
- Moves the integrity check of merkleized maps out of sign_psbt, and manages its state on the API side. Calling the PSBT key accessor methods without first verifying that keys are uniques and sorted would violate the security requirements of the Merkle tree commitments. By moving the invariant checks inside the API, and marking the _keys_are_sorted check as private, we make it harder to accidentally violate the invariant.
- Add accessor methods for the various PSBT fields, so that raw PSBT values validation doesn't mix with logic of the `sign_psbt` handler; this also greatly improves readability, as PSBT accesses become self documenting and more concise.
- Incidentally, improve validation of the global fallback locktime field, avoiding mixing errors (failing validation) and absence.

### DIFF
--- a/src/common/merkle.h
+++ b/src/common/merkle.h
@@ -1,6 +1,7 @@
 
 #pragma once
 
+#include <stdbool.h>
 #include <stdint.h>
 
 // TODO: RFC6962 defines the empty list hash as sha256(b''); while we're using 0 here. Should we
@@ -86,4 +87,11 @@ typedef struct {
     uint64_t size;
     uint8_t keys_root[32];
     uint8_t values_root[32];
+
+    // PRIVATE - managed only by the merkleized-map API (call_get_merkleized_map* /
+    // call_check_merkleized_map_sorted). Set to true once the keys tree has been verified to be
+    // lexicographically sorted (and therefore the keys are unique), which is the precondition for
+    // reading a value by key. The by-key readers assert this is set.
+    // Callers must not read or set it directly.
+    bool _keys_are_sorted;
 } merkleized_map_commitment_t;
--- a/src/handler/lib/check_merkle_tree_sorted.h
+++ b/src/handler/lib/check_merkle_tree_sorted.h
@@ -58,6 +58,9 @@ static inline int call_check_merkle_tree_sorted(dispatcher_context_t *dispatcher
  */
 static inline int call_check_merkleized_map_sorted(dispatcher_context_t *dispatcher_context,
                                                    merkleized_map_commitment_t *map) {
+    // The map is not yet validated; explicitly mark it as such
+    map->_keys_are_sorted = false;
+
     int ret = call_check_merkle_tree_sorted(dispatcher_context, map->keys_root, (size_t) map->size);
     if (ret >= 0) {
         map->_keys_are_sorted = true;
--- a/src/handler/lib/check_merkle_tree_sorted.h
+++ b/src/handler/lib/check_merkle_tree_sorted.h
@@ -45,3 +45,22 @@ static inline int call_check_merkle_tree_sorted(dispatcher_context_t *dispatcher
                                                        NULL,
                                                        NULL);
 }
+
+/**
+ * Validates a merkleized map commitment whose fields were populated directly (rather than obtained
+ * from call_get_merkleized_map): checks that its keys tree is lexicographically sorted, and on
+ * success marks the commitment as validated so that its values can be read by key.
+ *
+ * This is the counterpart of call_get_merkleized_map for maps that are not fetched from an outer
+ * Merkle tree of maps (e.g. the PSBT global map, whose commitment comes straight from the APDU).
+ *
+ * Returns 0 on success, or a negative number on failure.
+ */
+static inline int call_check_merkleized_map_sorted(dispatcher_context_t *dispatcher_context,
+                                                   merkleized_map_commitment_t *map) {
+    int ret = call_check_merkle_tree_sorted(dispatcher_context, map->keys_root, (size_t) map->size);
+    if (ret >= 0) {
+        map->_keys_are_sorted = true;
+    }
+    return ret;
+}
--- a/src/handler/lib/get_merkle_leaf_index.c
+++ b/src/handler/lib/get_merkle_leaf_index.c
@@ -1,6 +1,8 @@
 #include <string.h>
 #include <limits.h>
 
+#include "get_merkle_leaf_index.h"
+
 /* Local headers */
 #include "client_commands.h"
 #include "get_merkle_leaf_hash.h"
@@ -21,7 +23,8 @@ int call_get_merkle_leaf_index(dispatcher_context_t *dispatcher_context,
         SET_RESPONSE(dispatcher_context, request, sizeof(request), SW_INTERRUPTED_EXECUTION);
     }
     if (dispatcher_context->process_interruption(dispatcher_context) < 0) {
-        return -3;
+        PRINTF("Interrupted execution failed.\n");
+        return MERKLE_LEAF_ERROR;
     }
 
     uint8_t found;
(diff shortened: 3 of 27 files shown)


src/handler/sign_psbt/psbt_fields.h
```diff
@@ -0,0 +1,143 @@
+/*****************************************************************************
+ *   Ledger App Bitcoin.
+ *   (c) 2025, 2026 Ledger SAS.
```
iartemov-ledger: Why 2025?

src/handler/lib/get_merkle_leaf_index.c
```diff
@@ -30,27 +33,32 @@ int call_get_merkle_leaf_index(dispatcher_context_t *dispatcher_context,
...
 
     if (found != 0 && found != 1) {
-        return -2;
+        PRINTF("Invalid value for the 'found' flag.\n");
+        return MERKLE_LEAF_ERROR;
     }
 
     if (!found) {
-        return -3;
+        // The client claims the leaf is not in the tree; this is not verified.
```
iartemov-ledger, line 46: Missed  `PRINTF()` here?

src/common/merkle.h
```diff
@@ -86,4 +87,11 @@ typedef struct {
     uint64_t size;
     uint8_t keys_root[32];
     uint8_t values_root[32];
+
+    // PRIVATE - managed only by the merkleized-map API (call_get_merkleized_map* /
+    // call_check_merkleized_map_sorted). Set to true once the keys tree has been verified to be
+    // lexicographically sorted (and therefore the keys are unique), which is the precondition for
+    // reading a value by key. The by-key readers assert this is set.
+    // Callers must not read or set it directly.
+    bool _keys_are_sorted;
```
iartemov-ledger, line 96: Not exactly a C, but sooner a python style.
On the contrary C11 §7.1.3 reserves any single leading underscore for file-scope identifiers ;)

src/handler/lib/check_merkle_tree_sorted.h
```diff
@@ -44,4 +44,23 @@ static inline int call_check_merkle_tree_sorted(dispatcher_context_t *dispatcher
...
+ * success marks the commitment as validated so that its values can be read by key.
+ *
+ * This is the counterpart of call_get_merkleized_map for maps that are not fetched from an outer
+ * Merkle tree of maps (e.g. the PSBT global map, whose commitment comes straight from the APDU).
+ *
+ * Returns 0 on success, or a negative number on failure.
+ */
+static inline int call_check_merkleized_map_sorted(dispatcher_context_t *dispatcher_context,
+                                                   merkleized_map_commitment_t *map) {
+    int ret = call_check_merkle_tree_sorted(dispatcher_context, map->keys_root, (size_t) map->size);
```
iartemov-ledger, line 64: As everything is in stack should not we reset  `_keys_are_sorted` to false as in `call_get_merkleized_map_with_callback()` in `src/handler/lib/get_merkleized_map.c` below:
```
    // The map is not yet validated; explicitly mark it as such
    out_ptr->_keys_are_sorted = false;
```
?

src/handler/lib/map_value_status.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+/**
+ * Outcome of reading a value by key out of a merkleized map.
```
iartemov-ledger: Need to ask your friend to make concise comment ;)

src/handler/sign_psbt/txhashes.c
iartemov-ledger: > `txhashes.c:593` — `tmp[MAX(32, 8+1+MAX_PREVOUT_SCRIPTPUBKEY_LEN)]` in
>    `compute_sighash_segwitv1` is now oversized. The witness UTXO goes into the accessor's own
>    buffer; the largest remaining use is the 32-byte txid. Should be `tmp[32]`
Makes sense ?

src/handler/sign_psbt/psbt_fields.h
```diff
@@ -0,0 +1,143 @@
...
+ */
+psbt_field_status_t psbt_get_input_redeem_script(dispatcher_context_t *dc,
+                                                 const merkleized_map_commitment_t *input_map,
+                                                 uint8_t *out,
+                                                 size_t out_cap,
+                                                 size_t *out_len);
+
+/**
+ * PSBT_IN_WITNESS_UTXO amount: reads the witness UTXO and returns its 8-byte little-endian amount.
+ * Only requires the value to be at least 8 bytes; the full structural validation of the witness
```
iartemov-ledger: `psbt_fields.h:116-123` — the witness-UTXO accessor says "at least 8 bytes" but also silently caps at `8 + 1 + MAX_PREVOUT_SCRIPTPUBKEY_LEN`.

src/handler/sign_psbt/psbt_fields.c
```diff
@@ -0,0 +1,194 @@
...
+ * Reads a value of variable length (up to `out_cap` bytes) into `out`, writing its length to
+ * `*out_len`. A value longer than `out_cap` is reported as PSBT_FIELD_ERROR.
+ */
+static psbt_field_status_t read_var(dispatcher_context_t *dc,
+                                    const merkleized_map_commitment_t *map,
+                                    uint8_t key_type,
+                                    uint8_t *out,
+                                    size_t out_cap,
+                                    size_t *out_len) {
+    int res = call_get_merkleized_map_value(dc, map, &key_type, 1, out, out_cap);
```
iartemov-ledger, line 56: ### 🚨 W1. Buffer accessors may partially write the caller's buffer on `PSBT_FIELD_ERROR`

- **File**: `src/handler/sign_psbt/psbt_fields.c:46-61` (`read_var`), affecting
  `psbt_get_input_prevout_txid`, `psbt_get_input_redeem_script`, `psbt_get_output_script`
- **CWE**: CWE-457 (use of uninitialized variable), latent
- **APDU-reachable**: yes — **not currently exploitable**

`read_var` passes the caller's buffer straight to `call_get_merkleized_map_value`, and
`get_merkle_preimage.c:126-130` verifies the preimage hash only *after* streaming every byte into
that buffer. A host that returns data failing its Merkle check therefore leaves unverified bytes in
the caller's buffer, and `*out_len` is never assigned. Same for a short value in
`psbt_get_input_prevout_txid`: 20 bytes get written, then `read_fixed` returns ERROR.

The seven scalar accessors are immune — they decode via a local `raw[]` and assign `*out` only on
PRESENT. (The "value too long for the buffer" case is also safe: checked at
`get_merkle_preimage.c:49`, before any write.)

**Why not HIGH**: all four call sites `SEND_SW` and return immediately on non-PRESENT; none reads
the buffer or the length afterwards. This is a contract gap, not a live bug.

**Fix**: document the ERROR-path behaviour in `psbt_fields.h` for those three, and set
`*out_len = 0` on every non-PRESENT return of `read_var` so the length can never be stale.

iartemov-ledger (commented):
Just few minor comments.

src/handler/lib/get_merkle_leaf_index.c
bigspider: Key being absent can be a normal occurrence, so I don't think we need to log anything.

src/handler/sign_psbt/psbt_fields.h
bigspider: Fixed in 48cb176452a19581be83221df81fea6238e96d7b

src/handler/lib/check_merkle_tree_sorted.h
bigspider: Can't hurt - added in 48cb176452a19581be83221df81fea6238e96d7b.

src/handler/lib/map_value_status.h
bigspider: Done in 48cb176452a19581be83221df81fea6238e96d7b

src/handler/sign_psbt/txhashes.c
bigspider: Changed to 32 in 48cb176452a19581be83221df81fea6238e96d7b

src/handler/sign_psbt/psbt_fields.h
bigspider: Updated comment in 48cb176452a19581be83221df81fea6238e96d7b

src/handler/sign_psbt/psbt_fields.c
bigspider: Good idea to add as defense-in-depth. Done in cf24d6370260b92f90e0bc656b4c7ee969010afa.

A similar fix could be added to all the various Merkle-tree related functions - leaving it for a separate PR, as it's unrelated to the changes here.

iartemov-ledger (approved):
Cool!

PR #537, opened 2026-07-28
Fix presets with sighash flags in app playground

bigspider:
It didn't correctly manage 'negative fees' when the fee is actually not computable.

### DIFF
--- a/dev-tools/playground/presets.py
+++ b/dev-tools/playground/presets.py
@@ -212,6 +212,9 @@ def build_psbt_from_spec(policy, spec: PsbtSpec) -> PSBT:
     explicit_total = sum(o.amount for o in spec.outputs if o.amount is not None)
 
     if unbalanced:
+        if spec.fee != 0:
+            raise ValueError("fee must be 0 (or omitted) when a sighash doesn't commit to all amounts")
+
         # No fee to spread around: such a sighash leaves the transaction open, so
         # there is nothing for a remainder output to absorb.
         if remainder_indices:
@@ -342,6 +345,23 @@ POLICY_PRESETS: List[PolicyPreset] = [
             KeySpec("m/48'/1'/0'/2'", external_index=0),
         ],
     ),
+
+    PolicyPreset(
+        name="tr-miniscript-3key",
+        description=(
+            "Taproot miniscript: internal keypath"
+            " + external single-key"
+            " + external timelocked single-sig"
+        ),
+        wallet_name="TR miniscript 3key",
+        template="tr(@0/**,{pk(@1/**),and_v(v:pk(@2/<0;1>/*),older(52560))})",
+        keys=[
+            KeySpec("m/86'/1'/0'"),                    # device: key-path spend
+            KeySpec("m/86'/1'/0'", external_index=0),  # ext1: immediate script-path
+            KeySpec("m/86'/1'/0'", external_index=1),  # ext2: timelock script-path
+        ],
+    ),
+
 
     # --- musig2 -------------------------------------------------------------
     PolicyPreset(
--- a/dev-tools/playground/presets.py
+++ b/dev-tools/playground/presets.py
@@ -167,6 +167,11 @@ class PsbtSpec:
     small `fee` with a change remainder gives a net receive. If every output has
     an explicit amount there is no free variable, so `fee` must equal
     `total_in - sum(outputs)` exactly (else it's a mistake and we raise).
+
+    All of that assumes the transaction is complete, which holds only as long as
+    every input commits to all the amounts — i.e. uses SIGHASH_DEFAULT or
+    SIGHASH_ALL (or omits the field). If there are other sighash flags, every
+    output must set `amount` explicitly, and `fee` is ignored.
     """
     inputs: List[PsbtInputSpec]
     outputs: List[PsbtOutputSpec]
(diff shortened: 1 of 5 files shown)


dev-tools/playground/presets.py
```diff
@@ -181,19 +186,42 @@ def build_psbt_from_spec(policy, spec: PsbtSpec) -> PSBT:
...
+
     remainder_indices = [n for n, o in enumerate(spec.outputs) if o.amount is None]
     if len(remainder_indices) > 1:
         raise ValueError("at most one output may omit `amount` (the remainder)")
     explicit_total = sum(o.amount for o in spec.outputs if o.amount is not None)
 
-    if remainder_indices:
+    if unbalanced:
+        # No fee to spread around: such a sighash leaves the transaction open, so
+        # there is nothing for a remainder output to absorb.
```
iartemov-ledger, line 219: An explicit `fee` verification might make sense for future evolution:
```
if unbalanced:
    if spec.fee != 0:
        raise ValueError("fee must be 0 (or omitted) when a sighash doesn't commit to all amounts")
```

test_utils/txmaker.py
```diff
@@ -397,9 +397,22 @@ def fill_inout(wallet_policy: WalletPolicy, inout: Union[PartiallySignedInput, P
...
+def sighash_commits_to_all_amounts(sighash: Optional[int]) -> bool:
+    """True if `sighash` fixes both the input and the output set, so that a
+    transaction signed with it must balance. `None` means PSBT_IN_SIGHASH_TYPE is
+    omitted, i.e. the default, which does. Everything else (NONE or SINGLE, which
+    leave outputs open, and any type with ANYONECANPAY OR-ed on top, which leaves
+    the inputs open) does not."""
+    return sighash is None or sighash in _SIGHASH_TYPES_COMMITTING_TO_ALL
+
+
+def createPsbt(wallet_policy: WalletPolicy, input_amounts: List[int], output_amounts: List[int], output_is_change: List[bool], input_is_external: Optional[List[bool]] = None, input_sighashes: Optional[List[Optional[int]]] = None) -> PSBT:
```
iartemov-ledger, line 414: The function's signature is of 414 bytes. 
We should think about applying a python linter. 
In `app-boilerplate` and others they started using `ruff`.

dev-tools/playground/presets.py
```diff
@@ -167,6 +167,11 @@ class PsbtSpec:
     small `fee` with a change remainder gives a net receive. If every output has
```
iartemov-ledger, line 167: Would it make sense in this PR to address the 2nd Oscar's request - the "the positive taproot-miniscript cleartext case" :
```
diff --git a/dev-tools/playground/presets.py b/dev-tools/playground/presets.py
index f109730f..3bc0ffd0 100644
--- a/dev-tools/playground/presets.py
+++ b/dev-tools/playground/presets.py
@@ -343,6 +343,22 @@ POLICY_PRESETS: List[PolicyPreset] = [
         ],
     ),
 
+    PolicyPreset(
+        name="tr-miniscript-3key",
+        description=(
+            "Taproot 3-key miniscript (BIP-388 test vector): "
+            "key-path device, OR script-path pk (ext1), "
+            "OR script-path pk+timelock (ext2, 52560 blocks)"
+        ),
+        wallet_name="TR miniscript 3key",
+        template="tr(@0/**,{pk(@1/**),and_v(v:pk(@2/<0;1>/*),older(52560))})",
+        keys=[
+            KeySpec("m/86'/1'/0'"),                    # device: key-path spend
+            KeySpec("m/86'/1'/0'", external_index=0),  # ext1: immediate script-path
+            KeySpec("m/86'/1'/0'", external_index=1),  # ext2: timelock script-path
+        ],
+    ),
+
     # --- musig2 -------------------------------------------------------------
     PolicyPreset(
         name="musig2-keypath",

```
?

iartemov-ledger (dismissed):
Approved with few comments.

dev-tools/playground/presets.py
bigspider: Added in c7a4b7915349640fe3766731073ab970a8d00885.

dev-tools/playground/presets.py
bigspider: Added in c7a4b7915349640fe3766731073ab970a8d00885, and it actually surfaced a bug in test-utils that was fixed in 1d1342dc141df9a639688206dab3d7f6535b1189 (I'll leave addressing the unnecessary code duplication for a separate PR).

test_utils/txmaker.py
bigspider: 414 is the line number :stuck_out_tongue_closed_eyes: 

I'd leave that to a separate PR, the `test_utils` folder could definitely be cleaned up in various ways.

dev-tools/playground/presets.py
```diff
@@ -320,6 +346,23 @@ def needs_registration(self) -> bool:
         ],
     ),
 
+    PolicyPreset(
+        name="tr-miniscript-3key",
+        description=(
+            "Taproot miniscript: internal keypath"
+            "+ external single-key"
```
iartemov-ledger: 2 copy pasted `+` to remove ;)

iartemov-ledger (dismissed):
See the remark.

dev-tools/playground/presets.py
bigspider: Actually, I just wanted to split the string `"Taproot miniscript: internal keypath + external single-key + external timelocked single-sig"` in multiple lines, but I did forget some spaces. Fixed (and rebased).

I also added commit 9f57d1a4ce97c4750f3ea42e8e941b9e9282f969 so that long preset names are not truncated (unless they don't fit, that is instead handled by the renderer via CSS).

iartemov-ledger: Ah, so the `+`'s are intentional?

iartemov-ledger (approved):

dev-tools/playground/presets.py
bigspider: Yes, just to separate the three alternative spending conditions.

PR #538, opened 2026-07-31
Harden prevout tx parsing checks

bigspider:
Also additional integrity checks on Merkle proofs.

### DIFF
--- a/src/handler/lib/psbt_parse_rawtx.c
+++ b/src/handler/lib/psbt_parse_rawtx.c
@@ -21,13 +21,13 @@ struct parse_rawtx_state_s;  // forward declaration
 
 typedef struct {
     struct parse_rawtx_state_s *parent_state;  // subparsers can access parent's state
-    unsigned int scriptsig_size;               // max 10_000 bytes
-    unsigned int scriptsig_counter;            // counter of scriptsig bytes already received
+    unsigned int scriptsig_size;
+    unsigned int scriptsig_counter;  // counter of scriptsig bytes already received
 } parse_rawtxinput_state_t;
 
 typedef struct {
     struct parse_rawtx_state_s *parent_state;
-    unsigned int scriptpubkey_size;     // max 10_000 bytes
+    unsigned int scriptpubkey_size;
     unsigned int scriptpubkey_counter;  // counter of scriptpubkey bytes already received
 } parse_rawtxoutput_state_t;
 
@@ -165,7 +165,7 @@ static const parsing_step_t parse_rawtxinput_steps[] = {
     (parsing_step_t) parse_rawtxinput_sequence,
 };
 
-const int n_parse_rawtxinput_steps =
+const size_t n_parse_rawtxinput_steps =
     sizeof(parse_rawtxinput_steps) / sizeof(parse_rawtxinput_steps[0]);
 
 /*   PARSER FOR A RAWTX OUTPUT */
@@ -283,7 +283,7 @@ static const parsing_step_t parse_rawtxoutput_steps[] = {
     (parsing_step_t) parse_rawtxoutput_scriptpubkey,
 };
 
-const int n_parse_rawtxoutput_steps =
+const size_t n_parse_rawtxoutput_steps =
     sizeof(parse_rawtxoutput_steps) / sizeof(parse_rawtxoutput_steps[0]);
 
 /*   PARSER FOR A FULL RAWTX */
@@ -510,7 +510,7 @@ static const parsing_step_t parse_rawtx_steps[] = {(parsing_step_t) parse_rawtx_
                                                    (parsing_step_t) parse_rawtx_witnesses,
                                                    (parsing_step_t) parse_rawtx_locktime};
 
-const int n_parse_rawtx_steps = sizeof(parse_rawtx_steps) / sizeof(parse_rawtx_steps[0]);
+const size_t n_parse_rawtx_steps = sizeof(parse_rawtx_steps) / sizeof(parse_rawtx_steps[0]);
 
 static void cb_process_data(buffer_t *data, void *cb_state) {
     psbt_parse_rawtx_state_t *state = (psbt_parse_rawtx_state_t *) cb_state;
--- a/src/handler/lib/psbt_parse_rawtx.c
+++ b/src/handler/lib/psbt_parse_rawtx.c
@@ -125,11 +125,11 @@ static int parse_rawtxinput_scriptsig(parse_rawtxinput_state_t *state, buffer_t 
     uint8_t data[32];
 
     while (true) {
-        int remaining_len = state->scriptsig_size - state->scriptsig_counter;
+        size_t remaining_len = state->scriptsig_size - state->scriptsig_counter;
 
         // We read in chunks of at most 32 bytes, so that we can always interrupt with less than 32
         // unparsed bytes
-        int data_len = MIN(32, remaining_len);
+        size_t data_len = MIN(32, remaining_len);
 
         bool read_result = dbuffer_read_bytes(buffers, data, data_len);
         if (!read_result) {
@@ -193,6 +193,12 @@ static int parse_rawtxoutput_scriptpubkey_size(parse_rawtxoutput_state_t *state,
     uint64_t scriptpubkey_size;
     bool result = dbuffer_read_varint(buffers, &scriptpubkey_size);
     if (result) {
+        // Every output is streamed and hashed here, since that is what the txid is computed from;
+        // therefore, we have to tolerate sizes larger than MAX_PREVOUT_SCRIPTPUBKEY_LEN
+        if (scriptpubkey_size > UINT32_MAX) {
+            return -1;
+        }
+
         state->scriptpubkey_size = (unsigned int) scriptpubkey_size;
 
         crypto_hash_update_varint(&state->parent_state->hash_context->header, scriptpubkey_size);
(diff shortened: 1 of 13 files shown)


src/handler/lib/psbt_parse_rawtx.c
```diff
@@ -193,6 +193,10 @@ static int parse_rawtxoutput_scriptpubkey_size(parse_rawtxoutput_state_t *state,
     uint64_t scriptpubkey_size;
     bool result = dbuffer_read_varint(buffers, &scriptpubkey_size);
     if (result) {
+        if (scriptpubkey_size > MAX_PREVOUT_SCRIPTPUBKEY_LEN) {
```
iartemov-ledger: Got a remark from a friend of mine that seems to be plausible:

> This applies to every output of the prevout tx, but only the requested one gets
> copied into `vout_scriptpubkey`. The others just pass through the hash for the
> txid, so a prevout tx with an OP_RETURN output (83 bytes) would now be rejected
> even when we're spending a plain P2PKH — making those UTXOs unsignable.


He proposes following fix:
```
diff --git a/src/handler/lib/psbt_parse_rawtx.c b/src/handler/lib/psbt_parse_rawtx.c
index a1171afb..d7a107bc 100644
--- a/src/handler/lib/psbt_parse_rawtx.c
+++ b/src/handler/lib/psbt_parse_rawtx.c
@@ -193,7 +193,10 @@ static int parse_rawtxoutput_scriptpubkey_size(parse_rawtxoutput_state_t *state,
     uint64_t scriptpubkey_size;
     bool result = dbuffer_read_varint(buffers, &scriptpubkey_size);
     if (result) {
-        if (scriptpubkey_size > MAX_PREVOUT_SCRIPTPUBKEY_LEN) {
+        // Every output is streamed and hashed here, since that is what the txid is computed from;
+        // only the requested one is copied out, and bounded below. The size is therefore checked
+        // only against the (unsigned int) truncation of the assignment that follows.
+        if (scriptpubkey_size > UINT32_MAX) {
             return -1;
         }
 
@@ -204,6 +207,12 @@ static int parse_rawtxoutput_scriptpubkey_size(parse_rawtxoutput_state_t *state,
         if (state->parent_state->output_index != -1) {
             unsigned int relevant_output_index = (unsigned int) state->parent_state->output_index;
             if (state->parent_state->out_counter == relevant_output_index) {
+                if (scriptpubkey_size > MAX_PREVOUT_SCRIPTPUBKEY_LEN) {
+                    // the requested output's scriptPubKey must fit in parser_outputs; such an
+                    // output is not spendable anyway
+                    return -1;
+                }
+
                 state->parent_state->parser_outputs->vout_scriptpubkey_len =
                     (unsigned int) scriptpubkey_size;
             }
```

Attached mock dispatcher unit test show the regression
[unit-tests.txt](https://github.com/user-attachments/files/30580952/unit-tests.txt).

iartemov-ledger (commented):
We seems to have a functional regression.

src/handler/lib/psbt_parse_rawtx.c
bigspider: You're friend is clever!
Fixed and rebased.

src/handler/sign_psbt/txhashes.c
```diff
@@ -248,6 +248,12 @@ bool __attribute__((noinline)) compute_tx_hashes(dispatcher_context_t *dc,
                 return false;
             }
 
+            if (in_scriptPubKey_len > MAX_PREVOUT_SCRIPTPUBKEY_LEN) {
```
iartemov-ledger, line 251: Seems to be also plausible.

### W-01. The two added length checks sit *after* the copy they appear to guard

- **Severity**: 🚨 WARNING (defense-in-depth; not APDU-reachable today)
- **File**: `src/handler/sign_psbt/txhashes.c:251` and `src/handler/sign_psbt/sign_input.c:278`
- **APDU**: `SIGN_PSBT (0x04)` — both sites are on the normal signing path
- **CWE**: CWE-120 (buffer copy without checking input size) — ordering defect
- **Exploitability**: None currently; the protection is real but located elsewhere

Both checks added by `4260da50` run *after* `get_amount_scriptpubkey_from_psbt*()` returns:

```c
    if (0 > get_amount_scriptpubkey_from_psbt(dc, &ith_map, &in_amount,
                                              in_scriptPubKey, &in_scriptPubKey_len)) { ... }

    if (in_scriptPubKey_len > MAX_PREVOUT_SCRIPTPUBKEY_LEN) {   // txhashes.c:251
        // this should never happen
```

But the copy into the caller's buffer happens *inside* that helper, at
`amount_from_psbt.c:76`:

```c
    *scriptPubKey_len = parser_outputs.vout_scriptpubkey_len;
    memcpy(scriptPubKey, parser_outputs.vout_scriptpubkey, parser_outputs.vout_scriptpubkey_len);
```

In `txhashes.c` the destination is `uint8_t in_scriptPubKey[MAX_PREVOUT_SCRIPTPUBKEY_LEN]`
(`:239`) — 34 bytes. If the length could ever exceed 34, the stack write would already have
happened by the time line 251 executes; the check would report a problem it could not prevent.

**Why it is not reachable.** The requested output's length is bounded at parse time in
`parse_rawtxoutput_scriptpubkey_size()`, so `vout_scriptpubkey_len <= 34` always holds. The two
checks are therefore assertions, not guards — which is what their comments say.

**Why it still matters.** The only thing protecting the 34-byte buffer is a bound three call levels
away, in the parser. This branch is itself the demonstration of how easily that bound moves: the
working-tree change relocated it into the `output_index`-matching branch. Nothing at the call sites
would have caught it if that relocation had been wrong.

**Fix.** Move the bound to the single choke point — inside
`get_amount_scriptpubkey_from_psbt_nonwitness()`, before the `memcpy`:

```c
    if (parser_outputs.vout_scriptpubkey_len > MAX_PREVOUT_SCRIPTPUBKEY_LEN) {
        return -1;
    }
```

That covers all five call sites at once (`preprocess_inputs.c:183`, `:201`, `txhashes.c:242`,
`sign_input.c:268`, `:312`) and makes the two added checks genuinely redundant, so they can stay as
cheap assertions.

iartemov-ledger (commented):
2nd round.

src/handler/lib/psbt_parse_rawtx.c
```diff
@@ -193,13 +193,25 @@ static int parse_rawtxoutput_scriptpubkey_size(parse_rawtxoutput_state_t *state,
     uint64_t scriptpubkey_size;
     bool result = dbuffer_read_varint(buffers, &scriptpubkey_size);
     if (result) {
+        // Every output is streamed and hashed here, since that is what the txid is computed from;
+        // therefore, we have to tolerate sizes larger than MAX_PREVOUT_SCRIPTPUBKEY_LEN
+        if (scriptpubkey_size > UINT32_MAX) {
```
iartemov-ledger, line 204: Less important. 

### W-02. Varint narrowing is guarded for one length field and not its siblings

- **Severity**: 🚨 WARNING (correctness / asymmetry; **not exploitable**)
- **File**: `src/handler/lib/psbt_parse_rawtx.c:108, 341, 384, 448, 463`
- **APDU**: `SIGN_PSBT (0x04)`
- **CWE**: CWE-681 (incorrect conversion between numeric types)

`scriptpubkey_size` is now guarded against the `(unsigned int)` narrowing (`> UINT32_MAX → -1`,
`:196`). The sibling length fields still narrow silently:

| Field | Line |
|---|---|
| `scriptsig_size` | `:108` |
| `n_inputs` | `:341` |
| `n_outputs` | `:384` |
| `cur_wit_stack_elements` | `:448` |
| `cur_wit_elem_len` | `:463` |

In each case `crypto_hash_update_varint()` commits the **full 64-bit** value to the txid hash while
the parser then consumes the **truncated 32-bit** count, so the hashed byte stream can be made to
diverge from the parsed one.

**Not exploitable.** None of these fields indexes a buffer: scriptSig and witness bytes are hashed
and discarded (32-byte local, reads capped at 32), and the counts are only loop bounds. The resulting
txid is the hash of a byte string that is not a valid transaction, so the `expected_prevout_hash`
comparison in `preprocess_inputs.c` rejects it. On the two paths passing `expected_prevout_hash =
NULL` the desync can only produce a wrong scriptPubKey, i.e. an invalid signature — self-harm, not a
bypass. Consistent with `doc/integration.md:188-192`, which already documents that legacy prevout
amounts are not trustworthy because the signature algorithm does not commit to them.

**Pre-existing**, not introduced here. Raised because this branch touched all three chunked loops and
guarded one field only, and because `eb7d6bb5` removed the `// max 10_000 bytes` comments that
previously (if inaccurately) suggested these fields were bounded — so a reader now has no signal
either way. A one-line `> UINT32_MAX` guard on each, or a comment stating the fields are intentionally
unbounded because they are hash-only, would settle it.

iartemov-ledger (commented):
3rd round ;)

src/handler/sign_psbt/txhashes.c
iartemov-ledger: Do we still need the `scriptpubkey_len > MAX_PREVOUT_SCRIPTPUBKEY_LEN`  checks in other places rather than in `src/handler/sign_psbt/amount_from_psbt.c`?

bigspider: actually, `get_amount_scriptpubkey_from_psbt_nonwitness` already had the extra check, I added it on the previous iteration. I now added it also to `get_amount_scriptpubkey_from_psbt_witness`. (rebased again)

It's ok to keep redundant checks, it can't hurt.

src/handler/lib/psbt_parse_rawtx.c
bigspider: I saw this one, but I decided it's a _wontfix_. There is no way this can cause problems: none of these fields can have larger value in practice, and passing a larger value (or any sort of invalid transaction) achieves nothing.

iartemov-ledger (dismissed):

iartemov-ledger (approved):

PR #539, opened 2026-08-03
Fix stack exhaustion

bigspider:
Some situations could cause stack exhaustion in register_wallet.

This adds stricter limits, and saves memory with some refactorings to use the stack more efficiently.
List of changes:

- Count wrappers in the recursion depth
- Limit the nesting of `thresh` fragments to `MAX_THRESH_NESTING = 4`
- add `__attribute__((noinline))` in some places to optimize stack usage; refactor `handler_register_wallet` to make _noinlining_ possible
- rewrite `key_orderings_count` so that memory utilization is linear in `CT_MAX_KEYEXPRS`, instead of quadratic
- Limit maximum size of the threshold in `thresh` fragments to 24.

### DIFF
--- a/src/common/wallet.h
+++ b/src/common/wallet.h
@@ -57,6 +57,13 @@
 #define MAX_DESCRIPTOR_TEMPLATE_LENGTH \
     MAX(MAX_DESCRIPTOR_TEMPLATE_LENGTH_V1, MAX_DESCRIPTOR_TEMPLATE_LENGTH_V2)
 
+// As parse_script is recursive, we set a maximum reasonable recursion depth in order to avoid the
+// risk of stack exhaustion.
+// This depth is unlikely to be hit in practice.
+// Miniscript wrappers are counted as well: while they are not parsed recursively, they still
+// increase the depth of the parsed policy, which affects other recursive walkers.
+#define MAX_PARSE_SCRIPT_RECURSION_DEPTH 16
+
 // at most 92 bytes
 // wallet type (1 byte)
 // name length (1 byte)
--- a/src/common/wallet.h
+++ b/src/common/wallet.h
@@ -67,6 +67,14 @@
 // Maximum supported nesting of thresh operators
 #define MAX_THRESH_NESTING 4
 
+// Maximum supported value for n in a thresh miniscript operator (technical limitation).
+// It also bounds the stack used while analyzing a policy: the arrays of compute_thresh_ops() and
+// compute_thresh_stacksize() are proportional to it, and up to MAX_THRESH_NESTING of them are
+// alive at the same time due to recursion, therefore this ends up eating a substantial amount of
+// memory.
+// This limit is extremely unlikely to be hit in practice.
+#define MAX_N_IN_THRESH 24
+
 // at most 92 bytes
 // wallet type (1 byte)
 // name length (1 byte)
--- a/src/common/wallet.h
+++ b/src/common/wallet.h
@@ -64,6 +64,9 @@
 // increase the depth of the parsed policy, which affects other recursive walkers.
 #define MAX_PARSE_SCRIPT_RECURSION_DEPTH 16
 
+// Maximum supported nesting of thresh operators
+#define MAX_THRESH_NESTING 4
+
 // at most 92 bytes
 // wallet type (1 byte)
 // name length (1 byte)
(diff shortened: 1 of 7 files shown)


src/common/wallet.h
```diff
@@ -57,6 +57,24 @@
 #define MAX_DESCRIPTOR_TEMPLATE_LENGTH \
     MAX(MAX_DESCRIPTOR_TEMPLATE_LENGTH_V1, MAX_DESCRIPTOR_TEMPLATE_LENGTH_V2)
 
+// As parse_script is recursive, we set a maximum reasonable recursion depth in order to avoid the
+// risk of stack exhaustion.
+// This depth is unlikely to be hit in practice.
+// Miniscript wrappers are counted as well: while they are not parsed recursively, they still
+// increase the depth of the parsed policy, which affects other recursive walkers.
+#define MAX_PARSE_SCRIPT_RECURSION_DEPTH 16
```
iartemov-ledger, line 65: There are two separate depth limits in the app, in two different pieces of code, and they don't agree:
1. Here, in the parser - the part that reads the descriptor when one register a wallet - allows a policy up to `16` levels deep. Since `wsh(...)` uses one level and `pk(...)` another, that leaves room for `15` wrappers.
2. The script builder - the part that later turns the policy into an actual address - keeps a fixed array of `10` slots as it works through the policy, and gives up if it needs an eleventh. In practice that allows `9` wrappers.

Following "reducing" alignment can be made:
```
diff --git a/src/common/wallet.h b/src/common/wallet.h
index 4caad225..671dd25a 100644
--- a/src/common/wallet.h
+++ b/src/common/wallet.h
@@ -62,7 +62,11 @@
 // This depth is unlikely to be hit in practice.
 // Miniscript wrappers are counted as well: while they are not parsed recursively, they still
 // increase the depth of the parsed policy, which affects other recursive walkers.
-#define MAX_PARSE_SCRIPT_RECURSION_DEPTH 16
+//
+// The value must not exceed MAX_POLICY_DEPTH, the depth that get_wallet_script() is able to
+// process: accepting a deeper policy would mean registering a wallet whose address the app is then
+// unable to derive. A static assertion in policy.c enforces the relation.
+#define MAX_PARSE_SCRIPT_RECURSION_DEPTH 10
 
 // Maximum supported nesting of thresh operators
 #define MAX_THRESH_NESTING 4
diff --git a/src/handler/lib/policy.c b/src/handler/lib/policy.c
index 86788eb7..f36871d9 100644
--- a/src/handler/lib/policy.c
+++ b/src/handler/lib/policy.c
@@ -18,8 +18,13 @@
 #include "segwit_addr.h"
 #include "wallet.h"
 
+// Depth of the stack of nodes used to walk a policy while producing its script. Policies deeper
+// than this cannot be processed, therefore the parser must not accept them in the first place.
 #define MAX_POLICY_DEPTH 10
 
+_Static_assert(MAX_PARSE_SCRIPT_RECURSION_DEPTH <= MAX_POLICY_DEPTH,
+               "the parser would accept policies whose script cannot be produced");
+
 // The last opcode must be processed as a VERIFY flag
 #define PROCESSOR_FLAG_V 1
```
This would be better for the stack consumption, but we also may want on the contrary to increase `MAX_POLICY_DEPTH` to 16.
As with 10 we have 0 margin for the 58 descriptor templates used in the repo.
 Raising `MAX_POLICY_DEPTH` to `16` instead, keeping the parser at `16`, moves get_wallet_internal_script_hash from 176 B to 224 B and leaves get_wallet_script unchanged. So probably better.

unit-tests/test_wallet.c
```diff
@@ -300,7 +300,7 @@ static void test_parse_policy_tr_musig_keypath(void **state) {
     assert_int_equal(root->base.type, TOKEN_TR);
```
iartemov-ledger, line 300: We may want to add following 2 ragger non regression tests:

1. with a policy far deeper than its template must be refused, without crashing the app
2. the one that measures the stack used to register a wallet

[test_register_wallet.diff.txt](https://github.com/user-attachments/files/30694948/test_register_wallet.diff.txt)

iartemov-ledger (commented):
Just 2 minor comments.

unit-tests/test_wallet.c
bigspider: I added (a variation of) these tests in d61da7a00c54575005f68566b95dced960c054eb.

src/common/wallet.h
bigspider: In aa23d4363593f01ef14e0f1aa0e945a6b644ee32, I deleted `MAX_POLICY_DEPTH` since now its meaning is actually aligned with `MAX_PARSE_SCRIPT_RECURSION_DEPTH`.

iartemov-ledger (approved):

PR #540, opened 2026-08-04
Disable `musig` in key expressions of `musig_a`/`sortedmulti_a`

bigspider:
The parsing of such fragments is not correctly handled in the current state.
Therefore, we cleanly disable it until the parser is generalized.

### DIFF
--- a/src/common/wallet.c
+++ b/src/common/wallet.c
@@ -437,7 +437,7 @@ int parse_policy_map_key_info(buffer_t *buffer, policy_map_key_info_t *out, int 
  * - Single key index:
  *   - @IDX/**
  *   - @IDX/<M;N>/*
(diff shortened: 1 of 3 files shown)


iartemov-ledger (commented):
It's OK, but Claude making the review has come to following comment in wallet.h:401:

> The thresh comment is wrong and it's the dangerous kind of wrong. It says "pointer to array of exactly n pointers to child scripts" — the same wording used for the genuinely array-indexed multi keys, but scriptlist is a linked list. Someone trusting that comment and writing scriptlist[i] would recreate exactly the bug this PR fixes. Worth correcting while the context is fresh.

iartemov-ledger (approved):

PR #542, opened 2026-08-04
Add some missing error propagation in `musig.c`

bigspider:
The return value of `crypto_tr_lift_x` and `cpoint` was not consistently checked.

### DIFF
--- a/src/musig/musig.c
+++ b/src/musig/musig.c
@@ -204,7 +204,10 @@ int musig_key_agg(const plain_pk_t pubkeys[], size_t n_keys, musig_keyagg_contex
             return -1;
         }
 
-        point_add(&ctx->Q, &P, &ctx->Q);
+        if (CX_OK != point_add(&ctx->Q, &P, &ctx->Q)) {
+            PRINTF("Point addition failed in musig_key_agg\n");
+            return -1;
+        }
     }
 
     if (is_point_infinite(&ctx->Q)) {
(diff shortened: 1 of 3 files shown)


src/musig/musig.c
```diff
@@ -302,7 +308,10 @@ int musig_nonce_agg(const musig_pubnonce_t pubnonces[], size_t n_keys, musig_pub
                 PRINTF("Musig2 nonce aggregation: invalid contribution from cosigner %d\n", i);
                 return -i - 1;
             }
-            point_add(&R_j, &R_ij, &R_j);
+            if (CX_OK != point_add(&R_j, &R_ij, &R_j)) {
+                PRINTF("Point addition failed in musig_nonce_agg\n");
+                return -1;
```
iartemov-ledger: The comment in https://github.com/LedgerHQ/app-bitcoin/blob/musig-fixes/src/musig/musig.h#L113 says:

>  On error, `-i - 1` is returned if the
>  * nonce provided by the cosigner with index `i` is invalid, in order to allow blaming for a
>  * disruptive signer.

Here we're somehow attributing a different, arithmetic-type error to cosigner 0.

Maybe we should return something as:
```
return -(int) n_keys - 1;  // out of the -i-1 range: no signer to blame
```
?

src/musig/musig.c
```diff
@@ -201,7 +204,10 @@ int musig_key_agg(const plain_pk_t pubkeys[], size_t n_keys, musig_keyagg_contex
             return -1;
         }
 
-        point_add(&ctx->Q, &P, &ctx->Q);
+        if (CX_OK != point_add(&ctx->Q, &P, &ctx->Q)) {
```
iartemov-ledger, line 207: We may want to prefix by `WARN_UNUSED_RESULT` the most relevant functions to avoid ignoring the return value. it could give something as follows at compilation:
```
/app/src/musig/musig.c:311:13: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
  311 |             point_add(&R_j, &R_ij, &R_j);
      |             ^~~~~~~~~ ~~~~~~~~~~~~~~~~~
1 error generated.
```

iartemov-ledger (commented):
A couple of minor points. Does block the merge?

src/musig/musig.c
bigspider: Good point.

Actually, since all the points are already checked valid, failure should never happen. In any case, since it's iteration `i`, I think it makes sense to return -i - 1 like the error above.

I rebased with this change, plus:
- made the case explicit from `size_t` to `int` for `i`;
- fixed error message in the caller, which would print the wrong index

src/musig/musig.c
bigspider: Good point, and we should do it systematically across the codebase. Opening an issue.

iartemov-ledger: https://github.com/LedgerHQ/app-bitcoin/issues/543

iartemov-ledger (approved):

PR #546, opened 2026-08-13
Get rid of relative pointers in wallet policy AST

bigspider:
The structs used to store the parsed Abstract Syntax Tree of the descriptor template used relative pointers (2-byte size) in order to save memory.

The memory saving is no longer so important now that the codebase no longer targets Nano S, so it is not worth keeping the additional code complexity.

### DIFF
--- a/src/common/wallet.c
+++ b/src/common/wallet.c
@@ -18,6 +18,12 @@
 
 #include "../crypto.h"
 
+// The key expressions of multi() and friends are allocated one by one, but then accessed as an
+// array through policy_node_multisig_t::keys; that only works if buffer_alloc() lays them out
+// back-to-back, that is, if their size is a multiple of the alignment it pads to.
+_Static_assert(sizeof(policy_node_keyexpr_t) % 4 == 0,
+               "policy_node_keyexpr_t must be a multiple of the buffer_alloc() alignment");
+
 typedef struct {
     PolicyNodeType type;
     const char *name;
@@ -530,9 +536,9 @@ static int parse_keyexpr(buffer_t *in_buf,
         memcpy(key_indexes, keys, sizeof(uint16_t) * n_musig_keys);
 
         musig_info->n = n_musig_keys;
-        i_uint16(&musig_info->key_indexes, key_indexes);
+        musig_info->key_indexes = key_indexes;
 
-        i_musig_aggr_key_info(&out->m.musig_info, musig_info);
+        out->m.musig_info = musig_info;
     } else {
         return WITH_ERROR(-1, "Expected key expression starting with '@', or musig");
     }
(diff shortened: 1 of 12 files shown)


src/common/wallet.c
```diff
@@ -18,6 +18,12 @@
 
 #include "../crypto.h"
 
+// The key expressions of multi() and friends are allocated one by one, but then accessed as an
+// array through policy_node_multisig_t::keys; that only works if buffer_alloc() lays them out
+// back-to-back, that is, if their size is a multiple of the alignment it pads to.
+_Static_assert(sizeof(policy_node_keyexpr_t) % 4 == 0,
+               "policy_node_keyexpr_t must be a multiple of the buffer_alloc() alignment");
```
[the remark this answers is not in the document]
bigspider: Nah, it's fine

src/common/wallet.h
iartemov-ledger: Actually **140** (1+1+64+1+40+1+32).

iartemov-ledger: Actually **148** (1+1+64+9+32+9+32)

iartemov-ledger: This budgets key origin info at `2 + 8 + 3*12 = 46`, i.e. `3` derivation steps, while `MAX_BIP388_XPUB_DERIVATION_STEPS` is `8` (worst case `2 + 8 + 8*12 = 106`).

iartemov-ledger: Dead include - no need to supply `LEDGER_ASSERT` anymore.

iartemov-ledger (dismissed):
Just few remarks about cooments. 
+ stale rationale comment in [fuzzing/mock/wallet_model.c:114-116](vscode-webview://09n308lj5d9tomrr3rd3i1qelqetr39p219dhdfdro3cjrjl190a/fuzzing/mock/wallet_model.c#L114-L116).

src/common/wallet.h
bigspider: Good catch - if I remember correctly, the maximum length of the name was 16 bytes in an earlier version.

Fixed in afd42b6d3a7d11b26b0b6d527e73bfb4437134db

bigspider: Fixed in afd42b6d3a7d11b26b0b6d527e73bfb4437134db.

bigspider: Good point, and actually a legit bug (not hit in practice because usually people will use small numbers for derivations like `0` and `1`, not `1274392059`...)

Fixed in afd42b6d3a7d11b26b0b6d527e73bfb4437134db; the memory size increase at wallet registration shouldn't be problematic.

bigspider: Removed in afd42b6d3a7d11b26b0b6d527e73bfb4437134db

iartemov-ledger (approved):

PR #549, opened 2026-08-14
Clear the output buffer on failures in merkle-tree related functions

bigspider:
Leaving host-chosen bytes on the stack on failure cases is not a bug pe rse, but could make potential other bugs easier to exploit.

As a defense-in-depth measure, this hardens the various Merkle-tree related functionalities to wipe the output buffer in case of failure.

### DIFF
--- a/src/handler/lib/get_merkle_leaf_element.c
+++ b/src/handler/lib/get_merkle_leaf_element.c
@@ -1,3 +1,5 @@
+#include <string.h>
+
 #include "get_merkle_leaf_element.h"
 
 /* Local headers */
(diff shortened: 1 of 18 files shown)


iartemov-ledger (approved):
With 2 comments from Claude (which seem minor):

> - One change isn't pure hardening. `MAP_VALUE_ABSENT` now zeroes the buffer. `map_value_status.h` says callers applying a default must branch on `MAP_VALUE_ABSENT`, so a caller that pre-seeded a default and treated absence as "keep it" would silently get zeros. The PR asserts none does; I checked all four callers and that holds — but it's now a load-bearing contract that belongs in the header, not just a comment.
> - `read_var`'s own wipe is now redundant. `psbt_fields.c::read_var` already zeroed on absent-or-error and documented it — so #549 generalises an existing convention (a point in its favour), but every `ABSENT` lookup now zeroes twice. Tidiness only; buffers run ~43 bytes to a few hundred.

PR #556, opened 2026-08-25
Remove obsolete workaround on stack utilization on Nano X

bigspider:
It is no longer needed, and this reclaims about 2.5kb of available stack size for every handler except `sign_psbt`.

Also removes some unnecessary includes.


src/constants.h
```diff
@@ -57,13 +57,12 @@
...
 /**
  * Maximum number of external outputs handled simultaneously.
- * On the Nano X, the stack size is limited to 8K at the OS level,
- * so the stack consumption has to be limited as well.
  */
 #ifndef TARGET_ID
 #error "bolos_target.h must be included (TARGET_* constants unavailable)"
 #endif
 #ifdef TARGET_NANOX
+// Nano X has smaller stack size, so we keep this smaller.
```
iartemov-ledger: 1. `MAX_EXT_OUTPUT_SIMPLIFIED_NUMBER` sizes arrays inside `ui_validate_transaction_state_t`, which lives in the `g_ui_state` global union - it's `.bss` not stack. 
2.  Given that `.bss` just dropped 2528 bytes, it might be worth measuring whether 16 fits  (to generalize) - probably as a follow-up, not here.

Makefile
```diff
@@ -170,9 +170,9 @@ DEFINES   += HAVE_BOLOS_APP_STACK_CANARY
 
 # Estimated maximum stack usage.
 # It acts as a build-time check to protect global variables.
-# On the Nano X, the stack size is limited to 8K at the OS level.
+# On the Nano X, the RAM size is smaller, so we keep a smaller stack size.
 ifeq ($(TARGET_NAME),TARGET_NANOX)
-    APP_STACK_MIN_SIZE := 8192
+    APP_STACK_MIN_SIZE := 12288
```
iartemov-ledger, line 175: We got for NanoX:
- 28 672 of total SRAM
- 4 436 bytes used currently by `.bss`
- 11 097 bytes as worst case consumption for the stack

=> we should be able to set `APP_STACK_MIN_SIZE := 16384`  for all targets with 7 852 bytes margin for `.bss` to grow?

iartemov-ledger (commented):
At least the comment in `src/constants.h` has to be corrected.

src/constants.h
bigspider: [Updated the comment](https://github.com/LedgerHQ/app-bitcoin/compare/30a122a3c0e9e9df7570907c59042d2a9e1dcca2..38fb48f4b224aa2a1eb21cc391275f00f5b4ae5b) to mention RAM instead of stack size.

Makefile
bigspider: We probably could, but I'd leave that for a followup PR.

It might still make sense to keep a larger margin on Nano X by sacrificing a non-essential feature ("transactions sending to many different outputs") that is unlikely to me encountered by user - so that we have some available RAM on all devices for new features, without needing breaking changes later.

iartemov-ledger (approved):

PR #557, opened 2026-08-27
Various hardenings

bigspider:
Various hardenings across the codebase.


iartemov-ledger (approved):
`get_wallet_address` seems still to have a gap. 
It checks the HMAC, recomputes the wallet id, derives the address, and shows it - with no test that any key belongs to your device.

PR #558, opened 2026-08-28
Mark vendored embit clone 'private' in python client

bigspider:
It was always meant for internal use only, and not part of the public interface.


test_utils/musig2.py
```diff
@@ -46,7 +46,7 @@
     extract_placeholders,
 )
 
-from bitcoin_client.ledger_bitcoin.embit.descriptor.miniscript import Miniscript
+from bitcoin_client.ledger_bitcoin._embit.descriptor.miniscript import Miniscript
```
iartemov-ledger, line 49: The `test_utils`  itself violates the `_embit` "private" idea.

iartemov-ledger (approved):
Approving with a couple of ponits:
- should `ledger-bitcoin` version be incremented and `bitcoin_client/CHANGELOG.md`  be updated?
- would be the real fix to use https://github.com/diybitcoinhardware/embit/releases#release-v0.8.1 that seems to contain the "miniscript boolean literal support for 0 and 1" (not deployed on https://pypi.org/project/embit/#history)  instead of duplicating on our end?

test_utils/musig2.py
bigspider: Yeah, but it's tests, I think it's fine for now.

PR #567, opened 2026-09-17
Backport v2.5.1 to NanoS

iartemov-ledger:
### .text sizes:
- before:
```
round_to_2kb(62421) = 63488 bytes = 62.0 KB; diff = 1067 byte
```
- after:
```
round_to_2kb(62805) = 63488 bytes = 62.0 KB; diff = 683 bytes
```
=> still under 62k

No changes for RAM (bss/stack).

### Tested
On real Nano S device versus LW, made a successful TX.


bigspider (approved):
LGTM

PR #569, opened 2026-09-18
Client fixes on PSBTv2 field handling

bigspider:
Fixes various small issues in the client libraries handling of PSBT fields, and conversions between Psbt V0 and V2:

- Rust: the v0-to-v2 conversion didn't correctly take into account the unknown fields, that might be defined in the future. This also comes with some simplification, as the (de)serialization code is delegated to rust-bitcoin instead of re-implementing it here.  Bumped the minor version of the library, as the public members of the `psbt` module changed.
- Python: several small bugfixes related to Psbtv2 fields that were incorrectly managed. Patch version bumped.
- JS: fixed a hardcoded PSBT that had some invalid fields (currently ignored by the app, but this might change)


bitcoin_client/ledger_bitcoin/psbt.py
```diff
@@ -1051,7 +1051,7 @@ def cache_unsigned_tx_pieces(self) -> None:
         """
         # To make things easier, we split up the global transaction
         # and use the PSBTv2 fields for PSBTv0
-        if self.tx is not None:
+        if self.version == 0:
```
iartemov-ledger, line 1054: We might need a `bitcoin_client/tests/test_psbt.py` alongside `test_ripemd160.py` UT to catch such things.

iartemov-ledger (approved):

bitcoin_client/ledger_bitcoin/psbt.py
bigspider: I wish there was a decent python library for psbt, so we could delete this entirely (and only keep here the Merkleized map encoding machinery, which really is the only custom part).

